PR #3, opened 2020-01-14
Automation: report target data inconsistencies

acabarbaye:
### Description

determine target data inconsistencies and publish report daily on S3

### Test Coverage


ci_scripts/generate_news.py
```diff
@@ -1,64 +1,71 @@
...
+    RELEASE = 3
 
-    # how significant a change in version scheme should trigger a new changelog entry
-    # (api major, api minor, sdk major, sdk minor, sdk patch)
-    sigfigs = 4
-    current_version = LooseVersion(current).version
-    last_known_version = LooseVersion(last_known).version
 
-    should_towncrier = current_version[:sigfigs] != last_known_version[:sigfigs]
+def version(commit_type):
```
muchzill4: I'm not familiar with what `autoversion` does so it took me a while to grok what this is doing. I think it might read better if:
1. `version` was renamed to something more descriptive (it might be that it just doesn't _click_ for me) - currently its doing quite a bit (see 2.)
2. The functionality was separated into `calculate_version` and `write_news`.

Thoughts?

ci_scripts/generate_news.py
```diff
@@ -1,64 +1,71 @@
...
+                  CommitType.BETA: "prerelease"}
 
-    print('%s -- %s :: current vs previous changelog build' % (current, last_known))
-    if should_towncrier:
-        print('%s >> %s :: running changelog build' % (current, last_known))
+    old, new, updates = auto_version_tool.main(
+        release=commit_type == CommitType.RELEASE,
+        enable_file_triggers=True if commit_type in [CommitType.BETA,
+                                                     CommitType.RELEASE] else None,
+        bump=BUMP_TYPES.get(commit_type, None),
```
muchzill4: `{}.get()` returns `None` by default, so it's not necessary to pass `None` as `default` arg.

ci_scripts/generate_news.py
```diff
@@ -1,64 +1,71 @@
...
+        config_path=PROJECT_CONFIG,
+    )
+    logger.info('Version: %s' % new)
+    if commit_type in [CommitType.BETA,
```
muchzill4: This `in` check seems to be repeated, is it worth naming it something? `will_be_released =`? (somebody can probably come up with a better name)

ci_scripts/publish_target_validation_report.py
```diff
@@ -0,0 +1,35 @@
...
+
+def generate_validation_report_and_publish():
+    with tempfile.TemporaryDirectory() as tmp_directory:
+        platform_validator = PlatformValidator(tmp_directory, True)
+        platform_validator.render_results()
+        if platform_validator.processing_error:
+            e = Exception("Report generation failed")
+            logger.error(e)
+            raise e
+        upload_file(os.path.join(tmp_directory, 'index.html'), 'validation')
```
muchzill4: I'm not sure if the `PlatformValidator` interface is here to blame, but having implicit `index.html` as side effect is a bit smelly - if the output of `PlatformValidator` changes, this might end up being a hard to spot bug. I wonder if it wouldn't be more future proof to upload the entire contents of `tmp_directory` instead (i.e. not target `index.html` explicitly). 🤔

ci_scripts/publish_target_validation_report.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def main():
+    """Handle command line arguments to generate a results summary file."""
+
+    logging.basicConfig(level=logging.INFO,
```
muchzill4: How come `basicConfig` is required here, but not in the other command?

ci_scripts/aws_helpers.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+import mimetypes
+
+BUCKET = 'mbed-target'
+
+logger = logging.getLogger(__name__)
+
+S3_REGION = os.getenv('AWS_DEFAULT_REGION', 'eu-west-2')
+S3_CONFIG = {
+    "aws_access_key_id": os.getenv('AWS_ACCESS_KEY_ID', None),
```
muchzill4: `getenv` similarly to `{}.get` returns `None` by default if key does not exist.

ci_scripts/aws_helpers.py
```diff
@@ -0,0 +1,41 @@
...
+    "aws_secret_access_key": os.getenv('AWS_SECRET_ACCESS_KEY', None),
+}
+
+
+def upload_file(file, bucket_dir):
+    """
+    :param file: path to the file to upload
+    :param bucket_dir: name of the folder where to put the file in S3 bucket
+    """
+    logger.info('Uploading %s to AWS' % file)
```
muchzill4: Nit: `f'Uploading {file} to AWS'`. 👯‍♂

ci_scripts/publish_target_validation_report.py
acabarbaye: good point
Will do

acabarbaye:
Regarding `autoversion`, there is some documentation in this repository https://github.com/ARMmbed/autoversion but not much more. This will be improved in the future but did not manage to come around it just yet

ci_scripts/aws_helpers.py
muchzill4:
```suggestion
    logger.info(f'Uploading {file} to AWS')
```

azure-pipelines/build-release.yml
```diff
@@ -46,12 +47,13 @@ stages:
         versionSpec: '$(python.version)'
       displayName: 'Use Python $(python.version)'
 
-    # Note the code below generates a pip requirements file from the pipenv development requirements and then installs via
+    # Note the below generates a pip requirements file from the pipenv development requirements and then installs via
```
madchutney: The word `code` was also added in a later commit on my change which makes me worry as I don't think this is quite up to date with master. Also the list of commits seems wrong.

acabarbaye: the list of commits is due to the fact that this branch was started before your changes and was updated with your changes before you squashed and merged them to `master`.

acabarbaye:
> 
> 
> Thanks for applying suggested changes.
> 
> There's one more thing that I missed (🤦‍♂) when reviewing this PR in the morning. It is quite hard to review because of its size. It's much healthier to submit tiny PRs more frequently. The list of benefits is endless. 😉
> There's quite a lot happening here:
> 
>     * `DEVELOPMENT.md` documentation fixes
> 
>     * `requirements.txt/Pipfile` stuff
> 
>     * publishing target validation report
> 
>     * auto version + news = changelog
> 
>     * more?
> 
> 
> I know it's a bit late, but do you think it's worth doing something along the lines of:
> 
> ```
> git checkout -b generate-consistency-report
> git reset master
> git add azure-pipelines/build-release.yml ...
> git commit
> PR
> ```

I fully agree with the statement about small PRs when possible. However, for this particular one, it comprises work that was done over few sprints and which never got merged.
I am not sure I fully understand your latest suggestion regarding the `git` steps. Are you suggesting to split this PR into multiple PRs?

ci_scripts/generate_news.py
```diff
@@ -1,64 +1,80 @@
...
+    _generate_changelog(new_version, use_news_files)
+
+
+def _calculate_version(commit_type: CommitType, use_news_files: bool):
+    BUMP_TYPES = {CommitType.DEVELOPMENT: "build",
+                  CommitType.BETA: "prerelease"}
+    old, new_version, updates = auto_version_tool.main(
+        release=commit_type == CommitType.RELEASE,
+        enable_file_triggers=True if use_news_files else None,
+        bump=BUMP_TYPES.get(commit_type),
```
madchutney: From a code style I would prefer to see parameters calculated before they are passed into the function.

ci_scripts/generate_news.py
```diff
@@ -1,64 +1,80 @@
...
+"""
+
 import os
 import subprocess
+import logging
+import enum
+import argparse
+from auto_version import auto_version_tool
+from utils.definitions import NEWS_DIR, PROJECT_ROOT, PROJECT_CONFIG, \
+    LOGGER_FORMAT
```
madchutney: Is the line continuation needed with a 160 character limit?

ci_scripts/generate_news.py
```diff
@@ -1,64 +1,80 @@
...
-    current_version = LooseVersion(current).version
-    last_known_version = LooseVersion(last_known).version
-
-    should_towncrier = current_version[:sigfigs] != last_known_version[:sigfigs]
-
-    print('%s -- %s :: current vs previous changelog build' % (current, last_known))
-    if should_towncrier:
-        print('%s >> %s :: running changelog build' % (current, last_known))
+
+def version_project(commit_type: CommitType):
```
madchutney: 👍

ci_scripts/publish_target_validation_report.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def generate_validation_report_and_publish():
+    with tempfile.TemporaryDirectory() as tmp_directory:
+        platform_validator = PlatformValidator(tmp_directory, True)
+        platform_validator.render_results()
+        if platform_validator.processing_error:
+            e = Exception("Report generation failed")
+            logger.error(e)
+            raise e
```
madchutney: I would normally raise an exception and catch in the outside context if needed, and log there. Since this is a CI script I think it would be better to return a non-zero exit status (perhaps setting in the outside exception handler) to flag the CI step has failed rather than have the stack trace.

ci_scripts/aws_helpers.py
```diff
@@ -0,0 +1,41 @@
+import boto3
```
madchutney: Nice to have a top level docstring with a description.

ci_scripts/aws_helpers.py
```diff
@@ -0,0 +1,41 @@
...
+    :param file: path to the file to upload
+    :param bucket_dir: name of the folder where to put the file in S3 bucket
+    """
+    logger.info(f'Uploading {file} to AWS')
+    if not file or not os.path.exists(file):
+        raise FileNotFoundError(file)
+    client = boto3.client('s3', **S3_CONFIG)
+    dest_dir = (bucket_dir + '/' if bucket_dir else '')
+    dest_filename = os.path.basename(file)
+    key = '%s%s' % (dest_dir, dest_filename)
```
madchutney: It would be nice to use f strings as you have above.

ci_scripts/assert_news.py
```diff
@@ -1,86 +1,109 @@
...
-            )
-        )
-        exit(1)  # exit with an error status, no need for a traceback
-    print(':: %s new files in `%s`' % (len(added_news), news_dir))
-
-
-if __name__ == '__main__':
-    news_dir = sys.argv[1] if len(sys.argv) > 1 else None
-    main(news_dir)
+import logging
```
madchutney: It would be useful to have more comments in this file and use f strings if possible.

.gitignore
```diff
@@ -16,3 +16,5 @@ __pycache__/
 # pytest coverage
 .coverage
 htmlcov/
+
+dev-requirements.txt
```
muchzill4: Why is this required?

acabarbaye: CI generates this file and I wanted to ensure it never gets committed back to the repo

.gitignore
```diff
@@ -16,3 +16,5 @@ __pycache__/
 # pytest coverage
 .coverage
 htmlcov/
+
```
muchzill4:
```suggestion
```

.gitignore
muchzill4:
```suggestion
```

ci_scripts/utils/logging.py
```diff
@@ -0,0 +1,29 @@
...
+def set_log_level(verbose_count):
+    """
+    Sets the log level.
+    :param verbose_count: requested log level count
+    """
+    if verbose_count > 2:
+        log_level = logging.DEBUG
+    elif verbose_count > 1:
+        log_level = logging.INFO
+    elif verbose_count > 0:
```
muchzill4: I think it might be simpler if we did less greater than comparisons, it is a bit confusing otherwise (`> 0` should essentially only ever be `== 1` because of the condition above it).

What I mean:
```
if verbose_count > 2:
    log_level = logging.DEBUG
elif verbose_count == 2:
    log_level = logging.INFO
elif verbose_count == 1:
    log_level = logging.WARNING
else:
    log_level = logging.ERROR
```

ci_scripts/assert_news.py
```diff
@@ -1,86 +1,153 @@
...
+    def _verify_news_file_name(self):
+        basename = str(os.path.basename(
+            self.news_file_path)) if self.news_file_path else None
+        if (not basename) or (not re.search(NEWS_FILE_NAME_FORMAT, basename)):
+            raise ValueError(
+                f'Incorrect news file name [{basename}]. It must follow the following format: {NEWS_FILE_NAME_FORMAT}'
+            )
+        for extension in VALID_EXTENSIONS:
+            if extension in basename:
+                return
```
madchutney: If a break then else is used then then both validation rules follow the same flow and it also allows extra validation rules to be added in future following the same pattern.

```suggestion
                break
        else:
```

ci_scripts/assert_news.py
```diff
@@ -1,86 +1,153 @@
...
+
+VALID_EXTENSIONS = ['misc', 'doc', 'removal', 'bugfix', 'feature', 'major']
+NEWS_FILE_NAME_FORMAT = r"^[0-9]+.[a-z]+$"
+
+"""
+Ensures news file are created for all new changes to the project.
+"""
+
+
+class NewsFileValidator(object):
```
madchutney: There is no need to inherit from `object` in Python 3, it is implicit.
```suggestion
class NewsFileValidator:
```

ci_scripts/assert_news.py
```diff
@@ -1,86 +1,153 @@
...
+    """
+    Verification of the individual news files: naming, existence, content
+    """
+
+    def __init__(self, full_path):
+        self.news_file_path = full_path
```
madchutney: A general convention we should adopt is that if we don't expect the user to access the member variable we start the name with an underscore.
```suggestion
        self._news_file_path = full_path
```

ci_scripts/assert_news.py
```diff
@@ -1,86 +1,153 @@
...
+       Determines a list of all the news files which were added as part of the PR.
+        :return: list of introduced news files
+        """
+
+        if not os.path.exists(NEWS_DIR):
+            NotADirectoryError(
+                f'News files directory was not specified and default path `{NEWS_DIR}` does not exist'
+            )
+        logger.info(
+            f':: Looking for news files in `{NEWS_DIR}` [{self.current_branch}]'
```
madchutney: Is there a specific purpose of the leading `::`?

ci_scripts/assert_news.py
```diff
@@ -1,86 +1,153 @@
...
-
-if __name__ == '__main__':
-    news_dir = sys.argv[1] if len(sys.argv) > 1 else None
-    main(news_dir)
+import logging
+import os
+import re
+import sys
+import argparse
+from utils.definitions import PROJECT_ROOT, MASTER_BRANCH, BETA_BRANCH, \
```
madchutney: In the line continuation needed?

ci_scripts/assert_news.py
```diff
@@ -1,86 +1,153 @@
...
+
+    def verify(self):
+        """
+        Checks that news files were added in the current branch as part of the PR's changes
+        The files are then individually checked in order to ensure they follow the standard in terms of naming and content.
+        """
+        if self.current_branch in [MASTER_BRANCH, BETA_BRANCH] or re.search(
+                RELEASE_BRANCH_PATTERN, self.current_branch):
+            logger.info(
+                f'No need for news file on branch [{self.current_branch}]'
```
madchutney: Worth saying why there is no need for a news file, new readers may not have the background.

ci_scripts/assert_news.py
```diff
@@ -1,86 +1,153 @@
...
+            )
+            return
+        added_news = self.find_news_file()
+        if not added_news or len(added_news) == 0:
+            FileNotFoundError(
+                f'PR must contain a news file in {NEWS_DIR}. See README.md'
+            )
```
madchutney: Does this have an effect, did you mean to raise the exception?

ci_scripts/assert_news.py
```diff
@@ -1,86 +1,153 @@
...
+        description='Publish target data report to AWS.')
+    parser.add_argument("-v", "--verbose", action="count", default=0,
+                        help="Verbosity, by default errors are reported.")
+    args = parser.parse_args()
+    set_log_level(args.verbose)
+    try:
+        NewsFileDiscoverer().verify()
+    except Exception as e:
+        log_exception(logger, e)
+        sys.exit(1)
```
madchutney: It would be better to call sys.exit from where main is called, this may have undesired effects when called from elsewhere e.g. in tests
```suggestion
        return 1
    else:
        return 0
```

ci_scripts/assert_news.py
```diff
@@ -1,86 +1,153 @@
...
+
+
+if __name__ == '__main__':
+    main()
```
madchutney:
```suggestion
    sys.exit(main())
```

ci_scripts/generate_news.py
```diff
@@ -1,64 +1,86 @@
...
+    :type: CommitType
+    """
+
+    use_news_files = commit_type in [CommitType.BETA, CommitType.RELEASE]
+
+    new_version = _calculate_version(commit_type, use_news_files)
+    _generate_changelog(new_version, use_news_files)
+
+
+def _calculate_version(commit_type: CommitType, use_news_files: bool):
```
madchutney: Should we document the return type?

ci_scripts/generate_news.py
```diff
@@ -1,64 +1,86 @@
...
+    BUMP_TYPES = {CommitType.DEVELOPMENT: "build",
+                  CommitType.BETA: "prerelease"}
+    is_release = commit_type == CommitType.RELEASE
+    enable_file_triggers = True if use_news_files else None
```
madchutney: Not understanding how this work with the auto version tools, it might be useful to explain.

ci_scripts/generate_news.py
```diff
@@ -1,64 +1,86 @@
...
         )
-        with open(metafile, 'w') as fh:
-            fh.write(current)
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description='Versions the project.')
+    parser.add_argument('-r', '--release',
+                        help='Update the version for an official release',
```
madchutney:
```suggestion
                        help='Update the version for an production release',
```

ci_scripts/generate_news.py
```diff
@@ -1,64 +1,86 @@
...
+                        help='Update the version for a pre-release',
+                        action='store_true')
+    parser.add_argument("-v", "--verbose", action="count", default=0,
+                        help="Verbosity, by default errors are reported.")
+    args = parser.parse_args()
+    commit_type = CommitType.DEVELOPMENT
+    if args.release:
+        commit_type = CommitType.RELEASE
+    if args.beta:
+        commit_type = CommitType.BETA
```
madchutney: Would it be clearer to use a single argument with choices, something like:

`parser.add_argument('--release', default=CommitType.DEVELOPMENT, choices=[CommitType.RELEASE, CommitType.BETA, CommitType.DEVELOPMENT], help='...') 
`

ci_scripts/generate_news.py
```diff
@@ -1,64 +1,86 @@
...
+    set_log_level(args.verbose)
+
+    try:
+        version_project(commit_type)
+    except Exception as e:
+        log_exception(logger, e)
+        sys.exit(1)
```
madchutney: Same comment about calling sys.exit

ci_scripts/publish_target_validation_report.py
```diff
@@ -0,0 +1,42 @@
+import sys
```
madchutney: File level docstring would be nice

ci_scripts/publish_target_validation_report.py
```diff
@@ -0,0 +1,42 @@
...
+import tempfile
+from target_validation_report import PlatformValidator
+from utils.aws_helpers import upload_directory
+from utils.logging import log_exception, set_log_level
+
+logger = logging.getLogger(__name__)
+
+
+def generate_validation_report_and_publish():
+    with tempfile.TemporaryDirectory() as tmp_directory:
```
madchutney: Nice to have a summary docstring mentioning upload to AWS.

ci_scripts/publish_target_validation_report.py
```diff
@@ -0,0 +1,42 @@
...
+    set_log_level(args.verbose)
+
+    # Retrieve environment settings (if any) stored in a .env file
+    dotenv.load_dotenv(
+        dotenv.find_dotenv(usecwd=True, raise_error_if_not_found=False))
+    try:
+        generate_validation_report_and_publish()
+    except Exception as e:
+        log_exception(logger, e)
+        sys.exit(1)
```
madchutney: Same comment about calling sys.exit

ci_scripts/template.rst
```diff
@@ -0,0 +1,34 @@
+{# This is a custom template for towncrier md #}
```
madchutney: Could the filename give a hint that the template is for towncrier and markdown and not restructured text 'rst'?

ci_scripts/utils/aws_helpers.py
```diff
@@ -0,0 +1,74 @@
...
+logger = logging.getLogger(__name__)
+
+S3_REGION = os.getenv('AWS_DEFAULT_REGION', 'eu-west-2')
+S3_CONFIG = {
+    "aws_access_key_id": os.getenv('AWS_ACCESS_KEY_ID'),
+    "aws_secret_access_key": os.getenv('AWS_SECRET_ACCESS_KEY'),
+}
+
+
+def upload_file(file: str, bucket_dir: str):
```
madchutney: In this file it would be good to have some hints about why this is being done, knowing nothing about AWS I'm a bit lost here.

ci_scripts/utils/aws_helpers.py
```diff
@@ -0,0 +1,74 @@
...
+
+def upload_directory(dir: str, bucket_dir: str):
+    """
+    Upload the contents of a directory onto AWS S3
+    :param dir: folder to upload
+    :param bucket_dir: name of the folder where to put the directory contents in S3 bucket
+    """
+    logger.info(f'Uploading {dir} to AWS')
+    if not dir or not os.path.exists(dir):
+        raise FileNotFoundError(dir)
```
madchutney: Exception may be misleading if `dir` is not defined.

ci_scripts/utils/git_helpers.py
```diff
@@ -0,0 +1,122 @@
+import subprocess
+import os
+from .definitions import ENVVAR_GIT_TOKEN, REMOTE_ALIAS
+
+
+class GitWrapper(object):
+    def git_url_ssh_to_https(self, url: str):
```
madchutney: Document return type?

ci_scripts/utils/git_helpers.py
```diff
@@ -0,0 +1,122 @@
...
+            diff_command.append(dir)
+        return subprocess.check_output(diff_command).decode('utf-8')
+
+    def delete_branch(self, branch):
+        subprocess.check_call(['git', 'branch', '-D', branch])
+
+    def branch_exists(self, branch):
+        branches = subprocess.check_output(['git', 'branch']).decode(
+            'utf-8').strip()
+        return len([line for line in branches.splitlines() if
```
madchutney: This could probably do with some explanation

ci_scripts/utils/git_helpers.py
```diff
@@ -0,0 +1,122 @@
...
+                    line.replace('*', '').split() == branch.split()]) > 0
+
+    def get_changes_list(self, change_type, commit1, commit2, dir=None):
+        file_diff = self.get_changes(commit1, commit2, dir)
+        return [(line.strip()[len(change_type):]).strip() for line in
```
madchutney: I could do with some help here as to what is being done.

ci_scripts/utils/git_helpers.py
acabarbaye: part of #7

ci_scripts/utils/git_helpers.py
acabarbaye: part of #7

ci_scripts/utils/git_helpers.py
acabarbaye: part of #7

PR #4, opened 2020-01-15
Introducing types and enforcing checks as part of the CI

acabarbaye:
### Description
Use of python3 types in project source
Use of `mypy` to check their use in the project

This PR depends on #3 
### Test Coverage


mbed_targets/mbed_targets.py
```diff
@@ -1,53 +1,53 @@
...
-        self._features = self._attributes.get("features", {})
+        self._target_entry: dict = target_database_entry
+        self._attributes: dict = self._target_entry.get("attributes", {})
+        self._features: dict = self._attributes.get("features", {})
 
     @property
-    def board_type(self):
+    def board_type(self) -> str:
         """Compilation target.
         :rtype: str
```
madchutney: I think this kind of comment now superfluous?

Pipfile
```diff
@@ -6,9 +6,9 @@ verify_ssl = true
 [dev-packages]
 mock = "*"
 requests-mock = "*"
-jinja2 = "*"
```
rosiemiller2: How come jinja2 isn't needed anymore?

azure-pipelines/build-release.yml
```diff
@@ -62,7 +62,11 @@ stages:
     # Static analysis is different for Python versions so should be run in each environment.
     - script: |
         flake8
-      displayName: 'Static Analysis (flake8)'
+      displayName: 'Static Analysis - general (flake8)'
+
+    - script: |
+        mypy -p mbed_targets
+      displayName: 'Static Analysis - type checks (mypy)'
```
rosiemiller2: Something a little off with the indentation here.

Pipfile
acabarbaye: jinja2 is used by the target data inconsistencies report which is in /ci_scripts and is not needed by the actual mbed-target code

PR #5, opened 2020-01-17
🐛 Make setup.py actually install

muchzill4:
### Description

While exploring cli layout and cross-package dependencies I discovered that `mbed-targets` doesn't install correctly. At the moment `find_packages()` is called incorrectly with `"mbed_targets"` as argument, resulting in `_internal` being installed as package instead (`find_packages("mbed_targets") == ["_internal"]`):

```
❯ python setup.py install
...
❯ pip list
Package            Version
------------------ ----------
...
mbed-targets       0.1.0
...
```

```
❯ cd outside_of_project_directory # <- this is important, as otherwise mbed_targets is available as local package
❯ python
Python 3.7.3 (default, Oct 11 2019, 19:39:43)
[Clang 11.0.0 (clang-1100.0.33.12)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import mbed_targets
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'mbed_targets'
>>> import _internal
>>> _internal
<module '_internal' from '/Users/barmuc01/.local/share/virtualenvs/mbed-targets-39gjAcyB/lib/python3.7/site-packages/mbed_targets-0.1.0-py3.7.egg/_internal/__init__.py'>
```

This PR fixes the above issue, by swapping `find_packages()` with a single element list, which is preferred for simple projects.

I'm also using it as an opportunity to improve formatting of `setup.py` and fixing misc warnings.

### Test Coverage

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ with open(os.path.join(repository_dir, 'requirements.txt')) as fh:
 setup(
     author="Graham Hammond",
     author_email="support@mbed.com",
-    classifiers=(
+    classifiers=[
         'Development Status :: 3 - Alpha'
         'Intended Audience :: Developers',
         'License :: OSI Approved :: Apache Software License',
@@ -35,7 +35,7 @@ setup(
         'Programming Language :: Python :: 3.7',
         'Topic :: Software Development :: Build Tools'
         'Topic :: Software Development :: Embedded Systems',
-    ),
+    ],
     description="Mbed Tools: Target information for building Mbed OS",
     keywords="Arm Mbed OS MbedOS build target platform board module",
     include_package_data=True,
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,6 @@ setup(
     name=PACKAGE_NAME,
     packages=[SOURCE_DIR],
     python_requires='>=3.6,<4',
-    url="https://github.com/ARMmbed/%s" % PACKAGE_NAME,
+    url=f"https://github.com/ARMmbed/{PACKAGE_NAME}",
     version=__version__,
 )
--- a/setup.py
+++ b/setup.py
@@ -22,8 +22,8 @@ with open(os.path.join(repository_dir, 'requirements.txt')) as fh:
     requirements = fh.readlines()
 
 setup(
-    author="Graham Hammond",
-    author_email="support@mbed.com",
+    author='Graham Hammond',
+    author_email='support@mbed.com',
     classifiers=[
         'Development Status :: 3 - Alpha'
         'Intended Audience :: Developers',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -35,18 +34,17 @@
...
+    keywords='Arm Mbed OS MbedOS build target platform board module',
     include_package_data=True,
     install_requires=requirements,
     license='Apache 2.0',
     long_description_content_type='text/markdown',
     long_description=long_description,
     name=PACKAGE_NAME,
-    package_dir={'': SOURCE_DIR},
-    packages=find_packages(SOURCE_DIR),
+    packages=[SOURCE_DIR],
```
acabarbaye: what if we have subpackages within 'mbed_targets'?

setup.py
```diff
@@ -35,18 +34,17 @@
...
     python_requires='>=3.6,<4',
-    url="https://github.com/ARMmbed/%s" % PACKAGE_NAME,
+    url=f'https://github.com/ARMmbed/{PACKAGE_NAME}',
     version=__version__,
 )
```
acabarbaye: Could we also add entry points so that we don't need to always do `python -m ...`:
e.g.
    entry_points=dict(
        console_scripts=[
            "mbed_targets= mbed_targets.mbed_target:main",
            "mbedtargets = mbed_targets.mbed_target:main",
            "mbed-targets = mbed_targets.mbed_target:main",
            "mt = mbed_targets.mbed_target:main",
        ]

See https://setuptools.readthedocs.io/en/latest/setuptools.html
https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation
    ),

setup.py
muchzill4: Could you give me an example of such case? Unless we see an immediate need for it, I'd argue applying YAGNI rule for now. ;)

setup.py
muchzill4: If we were to do it, I'd rather put this in a separate PR.

- Do we really want this package to create console entry points though? My understanding was that there would be `mbed` (from `mbed-cli`) as an entry point? Like so https://click.palletsprojects.com/en/7.x/setuptools/ .
- What is `mbed_targets:mbed_target:main`? I can't find it.

acabarbaye: was an example of an entry point

setup.py
acabarbaye: any submodule that we want public. There is currently no example of this in mbed-targets but since we are using it as a template, I think it is worth considering that it may happen for other tools

setup.py
acabarbaye: I had the impression we would also be able to use the tools separately but I may have misunderstood after all. 
Since this PR is about fixing `setup.py`, I though we should tackle all use cases with regards to it but would understand if we consider it as good enough for now

setup.py
muchzill4: Resolving this for now as we took the discussion offline. Would like @madchutney to chip in as well.

I'd like to understand how `packages=` influences what's available. I understand it makes its contents available to import globally, but in our case:
```
find_packages(exclude=('tests',)) == ['mbed_targets', 'mbed_targets._internal']
```
 
What does `mbed_targets._internal` being there mean?

PR #8, opened 2020-01-20
✨ Mbed tools integration

muchzill4:
### Description

Simple integration point aligning with https://confluence.arm.com/display/IoTWeb/Sub-package+interface+for+presentation+of+command+line+options

### Test Coverage

- [X]  This change is covered by existing or additional automated tests.


mbed_targets/mbed_tools/cli.py
```diff
@@ -0,0 +1,5 @@
+from mbed_targets.mbed_tools.list_all import list_all as cli
```
rosiemiller2: I'm guess this will change when we want to add more than one command to the API? Seems a bit strange that `list_all == cli` but if it's just the example I think I get it.

rosiemiller2:
So is this the basic structure of the interfaces that we will need for the cookie cutter template? Does that mean we will have to remove the example stuff when we create a new repo from the template? Maybe I have misunderstood how cookie cutter works.

mbed_targets/mbed_tools/cli.py
muchzill4: It is just an example. :)

muchzill4:
I think this repo (`mbed-targets`) will stay in its current form. We're treating it as template/testing grounds for initial setup. I don't think it will be re-cookiecuttered, but I might be wrong. I.e.: it only serves as a blueprint of what we'll need the cookiecutter to contain and from then on we'll use cookiecutter to start new tools.

setup.cfg
```diff
@@ -21,6 +21,11 @@ per-file-ignores =
     mbed_targets/__init__.py:F401
 max-line-length = 160
 
+[coverage:run]
+omit =
+  # Top level entry point for mbed-tools, tested manually & in CI
+  mbed_targets/mbed_tools/cli.py
```
muchzill4: Could use some advice about this, as I'm not sure what's the best way to approach coverage report in files like this. It doesn't seem reasonable to exercise them by automated testing.

tests/mbed_tools/test_list_all.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@mock.patch("mbed_targets.mbed_tools.list_all.MbedTargets", spec_set=MbedTargets)
+def test_list_all_outputs_target_data(MbedTargets):
+    MbedTargets.return_value = [
+        mock.Mock(spec_set=MbedTarget, board_type="foo"),
+        mock.Mock(spec_set=MbedTarget, board_type="bar")
+    ]
+
+    runner = CliRunner()
```
rosiemiller2: This is cool.

tests/mbed_tools/test_cli.py
```diff
@@ -0,0 +1,6 @@
+from mbed_targets.mbed_tools.list_all import list_all
+from mbed_targets.mbed_tools.cli import cli
+
+
+def test_exposes_list_all_as_entry_point():
+    assert cli == list_all
```
acabarbaye: for tests, could we use unittest package and its helper method instead of assert?

muchzill4: FWIW, we had an offline discussion and @acabarbaye main concern was wrt to `self.assert*` family producing better error messages. (correct me if I'm wrong Adrien)

I'm not sure I agree with this, they both seem to have their flaws. Personally, I'm not a fan of verbosity of `unitttest.TestCase`. I much prefer succinct, yet not limiting `pytest` style where all you need is a function, but can go class if you so desire.
If we were to go with unittest style (unless it's already decided on) then I wonder why we use `pytest` in the first place. ;)
It's also worth mentioning, that you need workarounds for certain `unittest`y things with pytest (fixtures, `tearDown` when using `--pdb`).

madchutney: On balance I would prefer to stick with `unitttest.TestCase` and just use pyTest as a runner.

muchzill4: 👍 I shall update the test to use `unittest.TestCase` 😄

PR #12, opened 2020-01-21
Ensures the configuration of the project is in one place only

acabarbaye:
### Description

Removes duplication between toml file and definitions.py. The latter loads all the values it needs from the configuration file at load

### Test Coverage


ci_scripts/utils/definitions.py
```diff
@@ -1,14 +1,93 @@
...
+
+    """
+
+    class __ConfigLoader:
+        def __init__(self) -> None:
+            self.has_loaded: bool = False
+
+        def load(self) -> None:
+            """Loads values from the configuration file and sets the variables defined in this file with such values
+             """
```
madchutney: Indent mis aligned?

ci_scripts/utils/definitions.py
```diff
@@ -1,14 +1,93 @@
...
+
+
+class ConfigLoader:
+    """
+    Loads configuration values from `PROJECT_CONFIG` file
+    and sets the different variables.
+
+    """
+
+    class __ConfigLoader:
```
madchutney: `__` double underscore is normally reserved for system attributes.

ci_scripts/utils/definitions.py
```diff
@@ -1,14 +1,93 @@
...
+        def __init__(self) -> None:
+            self.has_loaded: bool = False
```
madchutney: Should this be an internal variable? For example:

```suggestion
            self._has_loaded: bool = False

```

ci_scripts/utils/definitions.py
```diff
@@ -1,14 +1,93 @@
...
+                    value = os.path.realpath(
+                        os.path.join(
+                            os.path.dirname(PROJECT_CONFIG),
+                            value
+                        )
+                    )
+                    break
+            return value
+
+    __instance = None
```
madchutney: I think this would be better at the top of the class, it makes it more obvious.

ci_scripts/utils/definitions.py
```diff
@@ -1,14 +1,93 @@
...
+
+    @staticmethod
+    def load():
+        """Loads definitions from project's configuration file
+        """
+        return ConfigLoader.getInstance().load()
+
+
+# Loads the configuration from the configuration file
+ConfigLoader.load()
```
madchutney: Do we really want this to execute this on an import?

muchzill4:
This might be unnecessarily complex. 🤔 Also, the fact that we need to implement and use a singleton pattern with`setattr` on the module rings alert bells in my head.

I think a more `flask`esuqe (python web framework) approach would work better here. I find it best explained by example:

```
# I separated "static" configuration for better visibility
class StaticConfig:
    VARIABLE_1 = True
    VARIABLE_2 = "something"

class Config(StaticConfig):
    def __init__(self):
        self._config_from_file = None

    def _get_value_from_config_file(self, name):
        if not self._config_from_file:
            self._config_from_file = self._load_config_from_file()
        return self._config_from_file.get(name, None)

    def _load_config_from_file(self):
        """Returns a dict containing k/v from config file"""
        return {
            'SOMETHING_FROM_FILE': 'rekt'
        }

    def __getattr__(self, name):
        """Load attributes not managed by config from the config file"""
        return self._get_value_from_config_file(name)


config = Config()

assert config.VARIABLE_1 is True
assert config.VARIABLE_2 == "something"
assert config.SOMETHING_FROM_FILE == "rekt"
```

And then in your module:

```
from something.configuration import config

config.VARIABLE_1
```

The downside here is that you cannot import the `VARIABLE_1` directly, but I think it's better than defining module variables dynamically. :) Thoughts?

acabarbaye:
Regarding @muchzill4 's review comments:
1. Having an OOP background, I find easier to use objects and abstract classes to define interfaces or common logic. I agree with you that it is not necessary but I find it easier to follow. I used the builtin abc.ABC though to follow best practices
2. Changed the constructor to take a list of sources instead
3. it is now lower case
4. I think we came to the conclusion that this resolution of path could be useful. So I left it.
5. I created an enumerator of the ones that need to be defined but as can be seen in `aws_helper` , you can use free text and it will be fine.

@rosiemiller2: I created a CLI tool to be able to get configuration information for external tools 
it is called `get_config.py` so in the future, we should be able to do something along the lines of : `$(get_config -k DOCUMENTATION_OUTPUT)`

ci_scripts/utils/configuration.py
```diff
@@ -0,0 +1,275 @@
...
+            self._found_file = True
+        except FileNotFoundError as e:
+            logger.warning(e)
+            self._found_file = False
+
+    def _find_config_file(self):
+        if self._found_file:
+            return
+        if self._file_path and os.path.exists(self._file_path):
+            self._found_file = True
```
muchzill4: I don't understand this bit. Why `self._found_file` is set to `True` and below in the `try` block it's set to `True` again? What's the case here?

This class has a bit too much internal state. I don't see how `self._found_file` and `self._loaded` would ever be needed on their own. The less your methods know about internal state and can operate on their own, the better. :)

How about something along the lines of:
```
class FileConfig(GenericConfig):
    def __init__(self, file_path: str = None) -> 'FileConfig':
        self._file_path: str = file_path
        self._config: Optional[dict] = None

    def _load_config_from_file(self):
        # orchestrate tree walking and trying to find the toml
        # ...
        return toml.load(f).get(FileConfig.CONFIG_SECTION, dict())

    def _get_config(self):
        # note, that this is the only place that cares if config was loaded or not,
        # no need for `_found_file` or `_has_loaded`
        if not self._config:
            self._config = self._load_config_from_file()
        return self._config

    def _fetch_value(self, key: str):
        config = self._get_config()
        try:
            value = config[key]
        except KeyError:
            self._raise_not_found(key)

        return self._make_path_absolute_to_toml(key, value)
```

ci_scripts/get_config.py
```diff
@@ -0,0 +1,37 @@
...
+import argparse
+import sys
+import logging
+from utils.configuration import configuration, ConfigurationVariable
+from utils.logging import set_log_level, log_exception
+
+logger = logging.getLogger(__name__)
+
+
+def main():
```
muchzill4: Happy to see argparsing happening in `main()`. It's one of the things we need to do, so that these files can be turned into executable scripts in the separate `-ci` repo, as `if __name__ == '__main__':` does not work there. 👍

acabarbaye: do I need to do anything here?

muchzill4: Nope, I'm just expressing my happiness.

acabarbaye:
Made the changes

ci_scripts/utils/configuration.py
acabarbaye: made the changes

ci_scripts/get_config.py
rosiemiller2: Ah, I haven't been doing that. I will change it in the docs generation script when I add in the release sync stuff.

ci_scripts/assert_news.py
```diff
@@ -99,12 +100,18 @@ def find_news_file(self):
         self.git.pull()
         master_branch_commit = self.git.get_current_commit()
         self.git.checkout(self.current_branch)
+        project_root = configuration.get_value(
+            ConfigurationVariable.PROJECT_ROOT)
```
madchutney: Style comment: Does this need to be on two lines?

ci_scripts/assert_news.py
```diff
@@ -99,12 +100,18 @@ def find_news_file(self):
...
+        news_dir_relative_path = os.path.relpath(news_dir,
+                                                 os.path.commonprefix(
+                                                     [news_dir, project_root]
+                                                 )
+                                                 )
```
madchutney: This section of code could do with a comment explaining what problem it is solving, it's not clear to me just be reading the code.

Although it is probably equivalent in this context I think `commonpath` would be what we would want, `commonprefix` is misleading if I understand the purpose of this section of code.

Lastly, would this be clearer if it was written with less indentation, e.g.

```
        news_dir_relative_path = os.path.relpath(
            news_dir, os.path.commonprefix([news_dir, project_root]))
```

ci_scripts/assert_news.py
```diff
@@ -141,17 +148,20 @@ def verify(self):
             )
             return
         added_news = self.find_news_file()
+        news_dir = configuration.get_value(ConfigurationVariable.NEWS_DIR)
+        project_root = configuration.get_value(
+            ConfigurationVariable.PROJECT_ROOT)
```
madchutney: I think this fits on one line

ci_scripts/get_config.py
```diff
@@ -0,0 +1,37 @@
...
+from utils.configuration import configuration, ConfigurationVariable
+from utils.logging import set_log_level, log_exception
+
+logger = logging.getLogger(__name__)
+
+
+def main() -> int:
+    """Handle command line arguments to get configuration values."""
+    parser = argparse.ArgumentParser(
+        description='Project\'s configuration.')
```
madchutney:
```suggestion
        description="Project's configuration.")
```

ci_scripts/utils/configuration.py
```diff
@@ -0,0 +1,276 @@
...
+        if not environment_value:
+            self._raise_undefined(key)
+        return environment_value
+
+
+class FileConfig(GenericConfig):
+    """Configuration set in toml file."""
+
+    CONFIG_SECTION = 'ProjectConfig'
+    PATH_TOKEN = {'DIR', 'ROOT', 'PATH'}
```
madchutney: This seems like special magic that isn't documented when adding configuration variables.

ci_scripts/utils/configuration.py
```diff
@@ -0,0 +1,276 @@
...
+        if not self._file_path:
+            return value
+        for token in FileConfig.PATH_TOKEN:
+            if token in variable_name:
+                value = os.path.realpath(
+                    os.path.join(
+                        os.path.dirname(self._file_path),
+                        value
+                    )
+                )
```
madchutney: I think this would be clearer if the full 160 character width was used.

ci_scripts/assert_news.py
acabarbaye: modifying this

ci_scripts/utils/configuration.py
acabarbaye: I wrote some more documentation about this behaviour

PR #13, opened 2020-01-21
Create build script for generating documentation

rosiemiller2:
### Description

Adding a script that will trigger a pdoc documentation build for mbed-targets. This can be hooked into our release scripts and the output files then synced to the repo.

### Test Coverage


ci_scripts/generate_docs.py
```diff
@@ -0,0 +1,27 @@
...
+    command_list = [
+        "pdoc", "--output-dir", f'{DOCUMENTATION_DIR_PATH}',
+        "--force", "--config", "show_type_annotations=True"
+    ]
+
+    for module_name in MODULES_TO_DOCUMENT:
+        command_list.append("--html")
+        command_list.append(f'{module_name}')
+
+    print('Creating pdoc documentation.')
```
acabarbaye: prefer `logging` package methods such as `info, warn, error`   for logging information as it gives more flexibility around where and what messages are shown to user

rosiemiller2: Sure 👍

ci_scripts/generate_docs.py
```diff
@@ -0,0 +1,30 @@
...
+logger = logging.getLogger(__name__)
+
+
+def main() -> None:
+    """Triggers building the documentation.
+
+    Modules to document and the output destination path
+    can be set in the utils.definitions config file.
+    """
+    command_list = [
```
acabarbaye: could the code be surrended by a `ty: except ` similar to this: https://github.com/ARMmbed/mbed-targets/blob/34fdfe05d6deed2cb26f243e597aa510a0586fc9/ci_scripts/publish_target_validation_report.py#L36?

(This was introduced because of Graham's point that if an exception is raised, we may not want the traceback to be shown)

rosiemiller2: Cool, I will wrap the `check_call` 👍

PR #17, opened 2020-01-22
📝Tweaks to docstrings

rosiemiller2:
### Description

Make sure they all follow the Google style docstring guidelines.
Add typing where it can be useful.

### Test Coverage


ci_scripts/requirements.txt
```diff
@@ -1,3 +1,6 @@
+#Requirements for the release process only
```
rosiemiller2: Is this correct guys? I made this mistake with the pdoc (ie. put it in here despite it being needed for every CI build) requirement so think its worth having some prompt here.

madchutney: I think a better description would be requirements for the CI process, as not all the scripts are related to release.

rosiemiller2:
@madchutney Yeah, I thought so too, but it looked to me like we don't install from that file when its not a beta or a production release? https://github.com/ARMmbed/mbed-targets/blob/master/azure-pipelines/build-release.yml#L52.

ci_scripts/requirements.txt
rosiemiller2: After talk IRL decided to change the ci_scripts directory structure to make this more obvious rather than rely on comments.

PR #20, opened 2020-01-22
📝 Document using code climate locally

muchzill4:
### Description

Documenting local usage of Code Climate.

### Test Coverage


DEVELOPMENT.md
```diff
@@ -95,3 +96,15 @@ relevant place using the [reST include](https://docutils.sourceforge.io/docs/ref
 Type hints should be used in the code wherever possible. Since the 
 documentation shows the function signatures with the type hints 
 there is no need to include additional type information in the docstrings.
+
+
+## Code Climate
+
+Code Climate is integrated with our GitHub flow. Failing the configured rules will yield a pull request not mergable.
+
+If you prefer to view Code Climate report on your machine, prior to sending a pull request you can use the [cli provided by Code Climate](https://docs.codeclimate.com/docs/command-line-interface)
```
rosiemiller2:
```suggestion
If you prefer to view the Code Climate report on your machine, prior to sending a pull request, you can use the [cli provided by Code Climate](https://docs.codeclimate.com/docs/command-line-interface).
```

DEVELOPMENT.md
```diff
@@ -95,3 +96,15 @@ relevant place using the [reST include](https://docutils.sourceforge.io/docs/ref
 Type hints should be used in the code wherever possible. Since the 
 documentation shows the function signatures with the type hints 
 there is no need to include additional type information in the docstrings.
+
+
+## Code Climate
+
+Code Climate is integrated with our GitHub flow. Failing the configured rules will yield a pull request not mergable.
```
rosiemiller2: Where are we configuring the rules? Aren't we just using CC out of the box?

muchzill4: We are using defaults, but they're kept in the repo in case we want to change something.

rosiemiller2: Do you reckon we could point people there so they can see what rules we use? The yml file isn't particularly readable, so maybe not :s

muchzill4: I think it's fine as is. If you end up doing something about the file, you'll need to read actual CC docs. Plus I'd rather keep only as much docs as necessary. :)

PR #21, opened 2020-01-22
👷 New CI scripts structure

rosiemiller2:
### Description

Because the current folder structure doesn't make it clear that the requirements in the requirements.txt file in the ci_scripts directory are only installed during release builds. This new structure should make that clearer.

This also separates out the docs generation script since we want that to be run on every build, release or not, and also locally.

### Test Coverage


azure-pipelines/consistency-report.yml
```diff
@@ -32,10 +32,10 @@ jobs:
 
       - script: |
           python -m pip install --upgrade pip
-          pip install -r ci_scripts/requirements.txt
+          pip install -r ci_scripts/release_scripts/requirements.txt
         displayName: 'Install requirements'
 
-      - script: python ci_scripts/publish_target_validation_report.py
+      - script: python ci_scripts/release_scripts/publish_target_validation_report.py
```
acabarbaye: - could we take this initiative to move this script to another folder like `validation_scripts` with its requirements.txt?
- Have also a `utils` folder with all shared scripts

rosiemiller2: Yep, that sounds like a great idea 👍

muchzill4:
Since this is an exceptional case of moving _bad_ files around you might need to manually approve them in Code Climate for it to pass. I hope it'll work. 😅

PR #22, opened 2020-01-22
licensing and Third-party IP report

acabarbaye:
### Description

Script determining all the project's third-party dependencies and generating SPDX files in the best way possible considering what's possible with the Python SPDX current SDK.

Note: Regarding OpenChain compliance, all the files will have to contain a header like:
```
 /*
  * This file is part of project X. It's copyrighted by the contributors
  * recorded in the version control history of the file, available from
  * its original location http://git.example.com/X/filename.c
  * 
  * SPDX-License-Identifier: Apache-2
  * License-Filename: LICENSE
  */
```
This will be done as part of another PR and potentially using https://github.com/johann-petrak/licenseheaders
### Test Coverage


acabarbaye:
> 
> 
> I tried running this locally, but it doesn't work. Did you forget to add dependencies in requirements file?
> 
> More and more I start to think we should have some sort of automated tests for these things. There's quite a lot of code here, which deals with a complex subject. It'll be really hard to do any sort of refactoring or bug fixing.
> 
> Also, files are cheap, I know the 250 line limit currently enforced by Code Climate might be arbitrary and not always applicable, but it is there for a reason. In my experience, splitting the code into small chunks - be it files or tiny classes, not only makes it easier to consume later on, but has added benefit of following single responsibility principle more.

To know the package dependencies, the script uses https://setuptools.readthedocs.io/en/latest/pkg_resources.html and has the same limitations as https://docs.python.org/3/library/importlib.metadata.html. Basically, the package you are trying to get the licence+3rd party info from needs to be installed. so you need to do the following
`python setup.py install`
then call the script
It is not great but this is the only way I found

rosiemiller2:
Has this been moved into the ci scripts repo?

acabarbaye:
Moved to `https://github.com/ARMmbed/mbed-tools-ci-scripts`

PR #23, opened 2020-01-23
Exclude target validation report

muchzill4:
Rationale being it's a throwaway script, to be replaced by the
`mbed-targets` in near future. Not worth investing time in refactor.

### Description

Exclude `target_validation_report.py` from Code Climate checks. It's temporary band aid, soon to be replaced with `mbed-targets`.

### Test Coverage

- [X]  Additional tests are not required for this change (e.g. documentation update).


.codeclimate.yml
```diff
@@ -30,3 +30,5 @@ checks:
   identical-code:
     config:
       threshold: # language-specific defaults. an override will affect all languages.
+exclude_patterns:
+  - "ci_scripts/target_validation_report.py"
```
madchutney: Does this need a path?

muchzill4: It does indeed. Fixed now. :)

PR #25, opened 2020-01-23
👷 Add configuration to generate docs in CI

rosiemiller2:
### Description

Run the docs build once every time a branch is pushed to origin. Move those docs to the artifacts folder in Azure pipelines. The documentation files can then be downloaded and checked.

### Test Coverage


azure-pipelines/build-release.yml
```diff
@@ -13,32 +13,32 @@ stages:
         Linux_Py_3_6:
           python.version: '3.6'
           vmImageName: ubuntu-18.04
-          uploadCoverage: "false"
+          triggerNonMatrixActions: "false"
```
rosiemiller2: Pretty terrible name - would welcome suggestions. Something that indicates that this is what triggers a step that we only want to run once, not on every machine in the matrix. :s

rosiemiller2: Maybe this is the wrong way to do it and we should have action specific attributes. eg. 
```
uploadCoverage: "true"
generateDocs: "true"
```

muchzill4: I like this approach (2 actions) better, it is more explicit and gives us more flexibility if we ever need it. I guess this is where I become a YAGNI hypocrite.

_Shame on me for suggesting single flag._

azure-pipelines/build-release.yml
```diff
@@ -83,6 +89,15 @@ stages:
...
 
+    - script: |
+        python ci_scripts/generate_docs.py --output_dir $(System.DefaultWorkingDirectory)/docs_output
+      displayName: 'Generate documentation'
+      condition: and(succeeded(), eq(variables['generateDocs'], 'true'))
+
+    - publish: $(System.DefaultWorkingDirectory)/docs_output
+      artifact: Documentation
+      condition: and(succeeded(), eq(variables['generateDocs'], 'true'))
+
```
madchutney: Would it make sense to do this in a separate stage? There is an overhead of installing the python dependencies but is this cleaner? The point of the `uploadCoverage` was that we needed the unit test results and didn't want to upload the results multiple times.

rosiemiller2: You are right I think it should be. I will set it up. 👍

rosiemiller2:
@madchutney This should be ready to go once it has finished building.

PR #26, opened 2020-01-23
Improves git add helper method

acabarbaye:
### Description

Makes git failure more explicit as well as the code wrapping the git command

### Test Coverage


ci_scripts/utils/git_helpers.py
```diff
@@ -49,13 +49,16 @@ def checkout(self, branch: Any) -> None:
         """
         self.repo.git.checkout(branch)
 
-    def _add_one_file(self, path: str) -> None:
+    def _add_one_file_or_one_dir(self, path: str) -> None:
+        if not path:
+            raise ValueError(f'Unspecified path: {path}')
```
rosiemiller2: Won't the path always be `None` if this is raised? Probably not worth putting that into the exception message.

acabarbaye:
```suggestion
            raise ValueError('Unspecified path.')
```

PR #27, opened 2020-01-23
💚️ Disable CI pipeline triggers which cause unnecessary builds.

madchutney:
### Description

Disable CI pipeline triggers which cause unnecessary builds.

### Test Coverage


setup.cfg
```diff
@@ -45,5 +45,5 @@ max-line-length = 160
 docstring-convention=google
 
 [tool:pytest]
-addopts = --cov mbed_targets --cov-report xml:coverage/coverage.xml
+addopts = --junitxml=junit/test-results.xml --cov mbed_targets --cov-report xml:coverage/coverage.xml --cov-report=html
```
muchzill4: Welp, I did remove `html` generation. Guess that'll be useful for local development. 👍 

Why is `junitxml` necessary?

muchzill4:
@madchutney Since this is OSS, I presume external contributions will come from forked repos. Is this configuration going to trigger builds on PRs that are from forks?

madchutney:
@muchzill4 That is a good question, forks are building fine now but I will need to test once I push to master. I suspect we will have to manually trigger the build of PRs from forks. So the question will be which of the following options we want:

1. Changes to a PR always trigger **two** builds except on forks when there will be **one**.
2. Changes to a PR always trigger **one** build except on forks where it will have to be **manually** triggered.

setup.cfg
madchutney: It was the recommended command line for Python, thought it might help with the warnings and be generally useful.

PR #29, opened 2020-01-24
💚️ Revise CI triggers for PRs and branch commits

madchutney:
### Description

- Trigger PRs when destination is master, beta or release branch.
- Trigger when code is merged to master, beta or release branch.
- Do not automatically build every branch.
- Disable Code Climate upload when secret is not available i.e. for PRs from forks.

### Test Coverage

Tested by creating PR from fork, also checked that Code Commit upload is performed when secret is available.


azure-pipelines/build-release.yml
```diff
@@ -82,13 +91,15 @@ stages:
...
+    # This script only runs for one set of unit tests (to avoid duplicate uploads) and will only run when the variable
+    # CODECLIMATE_REPORTER_ID is available, which means this step is disabled for PRs from forked repositories.
     - script: |
         curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
         chmod +x ./cc-test-reporter
         ./cc-test-reporter format-coverage -t coverage.py -o coverage/codeclimate.json coverage/coverage.xml
         ./cc-test-reporter upload-coverage
       displayName: 'Upload test coverage to Code Climate'
-      condition: and(succeeded(), eq(variables['uploadCoverage'], 'true'))
+      condition: and(succeeded(), eq(variables['uploadCoverage'], 'true'), ne(variables['CODECLIMATE_REPORTER_ID'], ''))
```
muchzill4: Is checking for `CODECLIMATE_REPORTER_ID` there in case somebody forks and sets up their own azure pipeline? Surely `uploadCoverage` should take care of duplicates, and in our environment `CODECLIMATE_REPORTER_ID` will always be set? 🤔

madchutney: No that isn't the reason. The variable is not available for PRs from forks as secrets are not exposed for security reasons. Therefore if we do not check if the variable is available then the build will fail, which is not what we want so better to skip this step.

PR #34, opened 2020-01-28
🔥 Use external ci scripts

muchzill4:
### Description

Extracting pipeline scripts to a separate repository. I want to stop here and keep this change small - i.e.: avoid moving `.toml` and doing additional `ci_scripts/` cleanup, as it's not strictly necessary for this to work.

Using an unreleased version for the time being, as I can't get them to actually release anything to PyPI. While this is not ideal, I believe its still worth merging, as:
- it unblocks this repository from potential conflicts and we can continue cleaning up `ci_scripts/` directory
- the `ci_scripts` are broken anyway, so having them elsewhere in similar state does not really change that much
- using PyPI released version vs direct github repo does not change that much at this stage

Extracted scripts repo: https://github.com/ARMmbed/mbed-tools-ci

### Test Coverage

- [X]  Additional tests are not required for this change (e.g. documentation update).


Pipfile
```diff
@@ -12,6 +12,7 @@ flake8-docstrings = "*"
 mypy = ">=0.500"
 pytest = "*"
 pytest-cov = "*"
-mbed-targets = {path = ".", editable = true}
 pdoc3 = "*"
 toml = "*"
+mbed-targets = {path = ".",editable = true}
```
rwalton-arm:
```suggestion
mbed-targets = {path = ".", editable = true}
```

muchzill4: I like this better, `pipenv` however does what I did by default for me. 🤔

PR #38, opened 2020-01-30
♻️ Refactor mbed_targets code.

rwalton-arm:
### Description

This patch refactors the mbed_targets code in the following areas:

* remove __next__ method from MbedTargets iterable container.
* Raise exceptions rather than returning None in
  _internal/target_database.py
* add some helper functions to _internal/target_database.py
* run black on mbed_targets source
* update tests for target_database.

### Test Coverage


tests/test_target_database.py
```diff
@@ -1,46 +1,39 @@
...
+        with pytest.raises(target_database.ResponseJSONError):
+            target_database.get_target_data()
+
+    def test_200_value_data(self, requests_mock):
         """Given a valid response, target data is set from the returned json."""
-        mock_request.get(target_database._TARGET_API, json={"data": 42})
+        requests_mock.get(target_database._TARGET_API, json={"data": 42})
         target_data = target_database.get_target_data()
-        self.assertEqual(42, target_data, "Target data is should match the contents of the target API data")
+        assert (
```
muchzill4: Not automatically enforced, but I think we should stick to `assertion` style from unittest where possible. 😞

mbed_targets/mbed_targets.py
```diff
@@ -53,22 +52,9 @@ class MbedTargets:
...
-
-        Returns:
-            An instance of a Mbed target database entry.
-        """
-        # Iterate over the target data and return a new MbedTarget instance for each database entry
-        return MbedTarget(
-            next(self._target_iterator)) if self._target_iterator else None
+    def __iter__(self) -> Iterator["MbedTarget"]:
+        """Yield an MbedTarget on each iteration."""
+        for tgt in self._target_data:
```
madchutney: I would prefer to be more verbose and not have to work out what tgt was shortened from.
```suggestion
        for target in self._target_data:
```

tests/test_target_database.py
rwalton-arm: Yeah I thought this may be controversial. My main reason for removing the `TestCase` inheritance was that it caused `pytest.mark.parametrize` to fail. We only parametrize one test here so I can revert to the `unittest` style if we must.

muchzill4: Well, I'd be on board with not using `TestCase` as it seems unnecessary most of the time. FWIW I did raise inability to use pytest fixtures as issue with `TestCase` in the past.
https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses

@acabarbaye @madchutney could we have some input please?

mbed_targets/_internal/target_database.py
```diff
@@ -2,41 +2,65 @@
...
 _TARGET_API = "https://os.mbed.com/api/v4/targets"
 
 
-def get_target_data() -> Optional[List[dict]]:
+class TargetAPIError(Exception):
+    """API request failed."""
+
+
+class ResponseJSONError(Exception):
+    """HTTP reponse JSON parsing failed."""
```
madchutney: This raises an interesting point, we should have a well defined exception hierarchy which is shared by all modules with a base class of a tools error so that it can be sensible caught by a calling script.

tests/test_target_database.py
```diff
@@ -1,46 +1,39 @@
...
-import unittest
-import requests_mock
+import pytest
 
 # Unit under test
 import mbed_targets._internal.target_database as target_database
 
 
-class TestGetTargetData(unittest.TestCase):
+class TestGetTargetData:
```
madchutney: Please stick to using unittest.TestCase, we have already had the discussion around this area and the conclusion was the test case assertions were more descriptive and the standard already used in Mbed OS.

mbed_targets/_internal/target_database.py
```diff
@@ -2,41 +2,65 @@
...
-    target_data = None
+        The target database as retrieved from the targets API
 
+    Raises:
+        ResponseJSONError: error decoding the reponse JSON.
+        TargetAPIError: error retrieving data from the Target API.
+    """
+    target_data: List[dict] = [{}]
     response = requests.get(_TARGET_API)
+    _check_response_status_code(response)
```
madchutney: I can understand the change in flow in performs the error handling in one place, but the multiple nested function calls have obfuscated what was a clear flow of control.

mbed_targets/mbed_targets.py
```diff
@@ -53,22 +52,9 @@ class MbedTargets:
 
     def __init__(self) -> None:
         """Retrieve target data from the online database."""
-        self._target_data: Optional[
-            List[dict]] = target_database.get_target_data()
+        self._target_data: List[dict] = target_database.get_target_data()
```
madchutney: This change exposes the current caller to the API of this package to the new exceptions without documentation.

mbed_targets/_internal/target_database.py
rwalton-arm: I disagree; the previous flow of control was quite hard to follow with the nested `try/except` and `if/else` statements; the congnitive complexity of the function in its previous form was too high for my liking (and code climate seemed to agree). I think this change makes the control flow easier to understand, generally flat is better than nested. Also using `check_` in the function name usually implies that an exception will be thrown if the function fails. I can add a comment to the `_check_response_status_code` to make things a bit clearer, though.

tests/test_target_database.py
rwalton-arm: Personally, I completely disagree that the `self.assert` style is clearer than using plain `assert` statements. Most of the python community also agrees these days. However, I don't want to rock the boat too much (I probably have already :trollface: ) - I will revert to the `unittest` style.

mbed_targets/mbed_targets.py
rwalton-arm: I'll add a comment to the class docstring to clarify which exceptions it can throw.

mbed_targets/_internal/target_database.py
rwalton-arm: Good point. I have created an exception hierarchy in the latest commit. Let me know what you think.

tests/test_target_database.py
rwalton-arm: I've moved back to `unittest` style in the latest commit. Let me know what you think.

tests/test_target_database.py
```diff
@@ -14,16 +14,16 @@ class TestGetTargetData(TestCase):
...
-            target_database._TARGET_API, status_code=401, text="Not Found"
+            target_database._TARGET_API, status_code=401
         )
         with pytest.raises(target_database.TargetAPIError):
             target_database.get_target_data()
 
     @requests_mock.mock()
     def test_html_404_error(self, mock_request):
-        """Given an HTML error code, TargetAPIError is raised."""
+        """Given a 404 HTML error code, TargetAPIError is raised."""
```
muchzill4:
```suggestion
        """Given a 404 HTTP error code, TargetAPIError is raised."""
```

tests/test_mbed_targets.py
```diff
@@ -57,12 +53,24 @@ def test_iterator(self, mocked_get_target_data):
...
+        self.assertEqual(
+            fake_target_data,
+            mbed_target_list,
+            "The list comprehension should match the fake data",
+        )
 
         # Iterate through the list checking the value returned matched the enumerated count
         for count, target in enumerate(mbed_targets):
-            self.assertEqual(count, target._target_entry["count"], "Iterator count values should match")
+            self.assertEqual(
```
muchzill4: I'm struggling to understand what this tests. :)

rwalton-arm: I didn't change any of the tests in this file, I just ran black on the code. I will look into this when we define the outward facing API, I suggest this is out of scope for this particular PR.

muchzill4: Agreed.

rwalton-arm:
I think I've addressed all of the review comments now. @madchutney could you please review again?

tests/test_target_database.py
```diff
@@ -1,46 +1,52 @@
...
-        """Given a 401 response, target data is not set."""
-        mock_request.get(target_database._TARGET_API, status_code=401)
-        target_data = target_database.get_target_data()
-        self.assertIsNone(target_data, "Target data is expected to be empty on an error")
+    def test_404(self, mock_request):
+        """Given a 404 error code, TargetAPIError is raised."""
+        mock_request.get(
+            target_database._TARGET_API, status_code=404, text="Not Found"
+        )
+        with pytest.raises(target_database.TargetAPIError):
```
madchutney: Please use https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

tests/test_target_database.py
```diff
@@ -1,46 +1,52 @@
...
-        """Given a 404 response, target data is not set."""
-        mock_request.get(target_database._TARGET_API, status_code=404, text="Not Found")
-        target_data = target_database.get_target_data()
-        self.assertIsNone(target_data, "Target data is expected to be empty on an error")
+    def test_401(self, mock_request):
+        """Given a 401 error code, TargetAPIError is raised."""
+        mock_request.get(
+            target_database._TARGET_API, status_code=401
+        )
+        with pytest.raises(target_database.TargetAPIError):
```
madchutney: Please use https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

tests/test_target_database.py
```diff
@@ -1,46 +1,52 @@
...
 
     @requests_mock.mock()
     def test_200_invalid_json(self, mock_request):
-        """Given a valid response but invalid json, target data is not set."""
-        mock_request.get(target_database._TARGET_API, text="invalid_json")
-        target_data = target_database.get_target_data()
-        self.assertIsNone(target_data, "Target data is expected to be empty on an error")
+        """Given a valid response but invalid json, JSONDecodeError is raised."""
+        mock_request.get(target_database._TARGET_API, json=None)
+        with pytest.raises(target_database.ResponseJSONError):
```
madchutney: Please use https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

tests/test_target_database.py
```diff
@@ -1,46 +1,52 @@
...
     @requests_mock.mock()
-    def test_200_no_data(self, mock_request):
-        """Given a valid response but no json data, target data is not set."""
-        mock_request.get(target_database._TARGET_API, json={})
-        target_data = target_database.get_target_data()
-        self.assertIsNone(target_data, "Target data is expected to be empty on an error")
+    def test_200_no_data_field(self, mock_request):
+        """Given a valid response but no data field, ResponseJSONError is raised."""
+        mock_request.get(target_database._TARGET_API, json={"notdata": []})
+        with pytest.raises(target_database.ResponseJSONError):
```
madchutney: Please use https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

tests/test_target_database.py
```diff
@@ -1,46 +1,52 @@
...
+            target_database.get_target_data()
 
     @requests_mock.mock()
     def test_200_value_data(self, mock_request):
         """Given a valid response, target data is set from the returned json."""
         mock_request.get(target_database._TARGET_API, json={"data": 42})
         target_data = target_database.get_target_data()
-        self.assertEqual(42, target_data, "Target data is should match the contents of the target API data")
+        self.assertTrue(target_data == 42, "Target data should match the contents of the target API data")
```
madchutney: Please use `assertEqual` as before

tests/test_target_database.py
```diff
@@ -1,46 +1,52 @@
 """Tests for `mbed_targets`."""
 
-import unittest
+from unittest import TestCase
+
+import pytest
```
madchutney: I don't think this is needed any longer

mbed_targets/mbed_targets.py
```diff
@@ -52,23 +55,17 @@ class MbedTargets:
...
+
+        Raises:
+            MbedTargetDataError: failed to get data from the target database.
         """
-        # Iterate over the target data and return a new MbedTarget instance for each database entry
-        return MbedTarget(
-            next(self._target_iterator)) if self._target_iterator else None
+        try:
+            self._target_data: List[dict] = target_database.get_target_data()
+        except target_database.TargetDatabaseError as err:
```
madchutney: I think we can let this exception go and just indicate that this method raises a  `ToolsError`

mbed_targets/_internal/target_database.py
madchutney: Lets chat about this.

mbed_targets/mbed_targets.py
rwalton-arm: I'd prefer not to catch such a general exception here. This class has a specific invariant, which is that we can actually get some data from the target database. Catching the `TargetDataBase` error communicates this to the caller. Catching general `ToolsError`s everywhere is almost as bad as catching a base `Exception`.

tests/test_target_database.py
rwalton-arm: 31fdc5e8dd760b77543315b2a32912061873f170

tests/test_target_database.py
rwalton-arm: 31fdc5e8dd760b77543315b2a32912061873f170

tests/test_target_database.py
rwalton-arm: I'm not sure there's a good reason for changing this to be honest, but have changed in 31fdc5e8dd760b77543315b2a32912061873f170

tests/test_target_database.py
rwalton-arm: 31fdc5e8dd760b77543315b2a32912061873f170

tests/test_target_database.py
rwalton-arm: 31fdc5e8dd760b77543315b2a32912061873f170

mbed_targets/_internal/target_database.py
```diff
@@ -1,42 +1,74 @@
...
-            try:
-                target_data = json_data["data"]
-            except KeyError:
-                logger.error(f"JSON received from '{_TARGET_API}' is missing the 'data' key.")
-    elif response.status_code == 401:
-        logger.error(f"Authentication failed for '{_TARGET_API}'. Please check that the environment variable "
-                     f"'{_AUTH_TOKEN_ENV_VAR}' is correctly configured with a private access token.")
-    else:
-        logger.error(f"An HTTP {response.status_code} was received from '{_TARGET_API}' containing:\n{response.text}")
+    if response.status_code != HTTPStatus.OK:
```
madchutney: I really like that, self descriptive and simple :)

mbed_targets/mbed_targets.py
madchutney: I meant we shouldn't catch the exception here at all, in the previous cases we caught a general exception and raised one based on `ToolsError` which is helpful to the user. I think whatever exception is raised in `get_target_data` can be passed back to the caller of this class. Raising a slightly different tools exception doesn't seem to add anything.

tests/test_target_database.py
```diff
@@ -1,46 +1,51 @@
...
-        """Given a 404 response, target data is not set."""
-        mock_request.get(target_database._TARGET_API, status_code=404, text="Not Found")
-        target_data = target_database.get_target_data()
-        self.assertIsNone(target_data, "Target data is expected to be empty on an error")
+    def test_401(self, mock_request):
+        """Given a 401 error code, TargetAPIError is raised."""
+        mock_request.get(
+            target_database._TARGET_API, status_code=401
+        )
+        with self.assertRaises(target_database.TargetAPIError):
```
madchutney: Thanks

tests/test_target_database.py
```diff
@@ -1,46 +1,51 @@
...
+            target_database.get_target_data()
 
     @requests_mock.mock()
     def test_200_invalid_json(self, mock_request):
-        """Given a valid response but invalid json, target data is not set."""
-        mock_request.get(target_database._TARGET_API, text="invalid_json")
-        target_data = target_database.get_target_data()
-        self.assertIsNone(target_data, "Target data is expected to be empty on an error")
+        """Given a valid response but invalid json, JSONDecodeError is raised."""
+        mock_request.get(target_database._TARGET_API, json=None)
```
madchutney: Is `json=None` actually invalid? I thought this was quite clear and guaranteed to be invalid JSON.

tests/test_target_database.py
```diff
@@ -1,46 +1,51 @@
...
+        with self.assertRaises(target_database.ResponseJSONError):
+            target_database.get_target_data()
 
     @requests_mock.mock()
     def test_200_value_data(self, mock_request):
         """Given a valid response, target data is set from the returned json."""
         mock_request.get(target_database._TARGET_API, json={"data": 42})
         target_data = target_database.get_target_data()
-        self.assertEqual(42, target_data, "Target data is should match the contents of the target API data")
+        self.assertEqual(target_data, 42, "Target data should match the contents of the target API data")
```
madchutney: The form of the assert methods are `(expected, actual, explanation)`
```suggestion
        self.assertEqual(42, target_data, "Target data should match the contents of the target API data")
```

mbed_targets/mbed_targets.py
rwalton-arm: Yeah, when you put it like that I think I agree, re-raising the slightly different `ToolsError` is a bit redundant. I'll make the change in the next commit.

tests/test_target_database.py
muchzill4: This is wrong and never explicitly stated in python docs.

1. `self.assertEqual(actual, expected)` reads better - "I expect server response to be this dict" vs "I expect this dict to be server response".
2. It is consistent with how you do conditionals, hence not an exception to memorize: `if foo == 2`, not `if 2 == foo`.
3. Observe:
```
    def test_dicts(self):
        # Result of calling subject under test
        subject = {
            'a': True,
            'b': False,
            'c': None,
        }
        self.assertEqual(
            subject, {
>               'd': set(),
            }
        )
E       AssertionError: {'a': True, 'b': False, 'c': None} != {'d': set()}
E       - {'a': True, 'b': False, 'c': None}
E       + {'d': set()}
```
The diff is true to the fact that lines with `-` are missing from my subject and lines with `+` are unexpected additions. Other way round it's just weird.

rwalton-arm: Yes, I agree - in the python docs it reads `(first, second)` and some linters advise to put the actual & expected the other way, too. Honestly didn't have the energy to push back on this as I thought it would spawn another 50 review comments

madchutney: @muchzill4 @rwalton-arm I agree that the documentation is far from clear but I base my request for change on the apparent implementation of `assertEqual`.

Changing the code to:

`self.assertEqual(41, target_data, "Target data should match the contents of the target API data")`

Gives the following warning:

```
Target data should match the contents of the target API data
42 != 41

Expected :41
Actual   :42
```

mark-edgeworth: @muchzill4 Looks like the conclusion from your example in point 3. above is the wrong way round :-). If 'd' is an unexpected addition then, by definition, 'subject' must be the thing that you are expecting to see. Hence the expected thing goes first in the call, and the actual thing goes second. This is the same as in jUnit, but differs from cUnit. Aren't standards wonderful :-)

muchzill4: @madchutney did you add `Expected` and `Actual`? I'm not able to make `pytest` nor `unittest` yield such output.

@mark-edgeworth I don't follow. :) Anyway, the more I look at 3., the more I realize it is fragile as it could be argued both ways, so you're right in that it's a moot point.

muchzill4: FWIW I had a browse in the `cpython` and the tests for `unittest` itself use `actual, expected`. 😈

madchutney: @muchzill4 I just ran them in PyCharm. Personally I don't like the ordering, but 🤷‍♂ it is what it is.

PR #43, opened 2020-02-04
Support target API authorisation token

rwalton-arm:
### Description

Pass an authorization header to `requests.get` if the `MBED_API_AUTH_TOKEN` env var is set.
Allow users to configure the API endpoint URL via an `MBED_TARGET_API_URL` environment variable.

### Test Coverage


mbed_targets/_internal/target_database.py
```diff
@@ -1,15 +1,34 @@
...
+import dotenv
 import requests
 
 
+# Search for the .env file containing the MBED_API_AUTH_TOKEN environment variable.
+# We want this to execute at import time.
+dotenv.load_dotenv(dotenv.find_dotenv(usecwd=True), override=True)
+
+
+def _set_target_api_url() -> str:
```
madchutney: I'm trying to think of a use case where a user would want to change this, but I haven't so far. If we can't think of a use case I don't think we should make it configurable.

mbed_targets/_internal/target_database.py
```diff
@@ -72,3 +90,14 @@ def _response_error_code_to_str(response: requests.Response) -> str:
         )
     else:
         return f"An HTTP {response.status_code} was received from '{_TARGET_API}' containing:\n{response.text}"
+
+
+def _get_request() -> requests.Response:
+    """Make a get request to the API, ensuring the correct headers are set."""
+    token = os.getenv(_AUTH_TOKEN_ENV_VAR)
+    if not token:
```
madchutney: I think it is simpler to to invert the logic of this if statement and remove the `not`.

mbed_targets/_internal/target_database.py
mark-edgeworth: Surely this should be '_get, not '_set'? Also looks a bit over-engineered to me; YAGNI?

mark-edgeworth: Also, why not os.getenv("MBED_TARGET_API_URL", "https://os.mbed.com/api/v4/targets")? Forgive my python noob question here.

mbed_targets/_internal/target_database.py
mark-edgeworth: Agree; flip that logic!

mbed_targets/_internal/target_database.py
```diff
@@ -95,9 +83,9 @@ def _response_error_code_to_str(response: requests.Response) -> str:
...
     token = os.getenv(_AUTH_TOKEN_ENV_VAR)
-    if not token:
-        header = None
-    else:
+    if token:
         header = {"Authorization": f"Bearer {token}"}
+    else:
+        header = None
 
     return requests.get(_TARGET_API, headers=header,)
```
mark-edgeworth: Could remove the comma here.

rwalton-arm: I have removed it in https://github.com/ARMmbed/mbed-targets/pull/45

`mypy` started failing after the suggestion to invert the logic - that's also fixed in the linked PR

rwalton-arm: Also, this is another reason we should use `black` in the CI.

PR #45, opened 2020-02-04
✨ Add a method to lookup a target by its product code

rwalton-arm:
### Description

Add a method to lookup a target by its product code
Also fix a mypy error in `target_database.py`

### Test Coverage


mbed_targets/mbed_targets.py
```diff
@@ -59,3 +59,21 @@ def __iter__(self) -> Iterator["MbedTarget"]:
...
+
+        Raises:
+            ValueError: the given product code was not found in the target database.
+        """
+        try:
+            return next(
+                target for target in self if target.product_code == product_code
+            )
+        except StopIteration:
+            raise ValueError(
```
madchutney: Would this be better as a tools exception, perhaps `UnknownTarget`? I could imagine this could be a common exception.

mbed_targets/_internal/target_database.py
```diff
@@ -83,9 +83,8 @@ def _response_error_code_to_str(response: requests.Response) -> str:
 def _get_request() -> requests.Response:
     """Make a get request to the API, ensuring the correct headers are set."""
     token = os.getenv(_AUTH_TOKEN_ENV_VAR)
+    header: Optional[Dict[str, str]] = None
```
muchzill4: Do you think adding test coverage for the header usage / construction from the env would be beneficial here?

muchzill4:
Gotta say, I love the size of this PR! 😍

mbed_targets/_internal/target_database.py
rwalton-arm: Yeah :) I added two tests for the header usage in the last PR. Here I'm only fixing a `mypy` error. The tests are in https://github.com/ARMmbed/mbed-targets/blob/master/tests/test_target_database.py

PR #46, opened 2020-02-05
🐛 Reinstate support for next() method on MbedTargets iterator

madchutney:
### Description

After checking the general consensus on the implementation of iterators I have reinstated support for calling using the `next()` method. While this is not the most likely use case, this is the least surprising to a caller. 

The other alternative is to remove the iterator functionality from the class and replace with a generator method which can be explicitly called. However, these seems more clumsy.

### Test Coverage


mbed_targets/mbed_targets.py
```diff
@@ -58,11 +58,25 @@ def __init__(self) -> None:
...
+
+    def __next__(self) -> "MbedTarget":
+        """Return the next item from the iterable.
+
+        Iterate over the target data and return a new MbedTarget instance for each database entry
+
+        Returns:
+            An instance of a Mbed target database entry.
+        """
+        return MbedTarget(next(self._target_iterator))
```
muchzill4: Is it desired that this re-uses `self._target_iterator` which might be reset by using `__iter__`? I rarely ever use `next` directly, so I'm not sure what this should do.

Example:
```
targets = MbedTargets()
t1 = next(targets) # has id = 1
for target in targets:
   t2 = target # has id = 1
   break
t3 = next(target) # has id = 2
```

mbed_targets/mbed_targets.py
```diff
@@ -58,11 +58,25 @@ def __init__(self) -> None:
             TargetDatabaseError: Failed to get the target data.
         """
         self._target_data: List[dict] = target_database.get_target_data()
+        self._target_iterator: Iterator = iter(self._target_data)
```
muchzill4: This `self._target_iterator` is odd, as it's overwritten in `__iter__`, yet reused in `__next__`. I find this a bit confusing to follow.

tests/test_mbed_targets.py
```diff
@@ -91,3 +106,6 @@ def test_lookup_by_product_code(self, mocked_get_target_data):
         mbed_targets = MbedTargets()
         target = mbed_targets.get_target(expected_product_code)
         self.assertEqual(expected_product_code, target.product_code, "Target's product code should match the given product code.")
+
+        with self.assertRaises(UnknownTarget):
```
muchzill4: Nice catch, but it might be better to test different code paths in separate `test_*`.

mbed_targets/mbed_targets.py
rwalton-arm: I think this is conceptually wrong. This `MbedTargets` class should be an iterable and not an iterator. We're actually abusing the iterator protocol here just so we can use `next` directly which is an unusual way to access elements in an iterable anyway.

rwalton-arm: Let me provide some rationale as to why this is an abuse of the iterator protocol
* Iterators are expected to be single pass
* Iterators don't contain data, they are supposed to just implement a `__next__` method which describes how to get the next element in an iterable. Having the iterable and iterator logic in the same class violates separation of concerns and makes the code much harder to understand.
 On the other hand, iterables:
* expected to contain data and can be iterated over
* expected be multi-pass
* return a new iterator on each call to their `__iter__` method

`next` is a function designed to be used with iterators and not iterable containers

Implementing `__iter__` as  a generator (what we had before this change) fulfilled the iterable protocol; generator objects are iterators, generator functions return generator objects. This class is now trying to be both a iterable and an iterator which will actually confuse users rather than giving them what they expect, and we've now made the code a lot harder to maintain..

PR #47, opened 2020-02-05
Don't have MbedTargets iterable

rosiemiller2:
Store the target data in a dict for easier lookups.

### Description
I'm not sure that having `MbedTargets` as an iterator makes sense if we are going to be looking up targets by product code. Iterating through the target list (and creating MbedTarget objects) on every `get_target` seems a bit inefficient. 

This way, we create and store a dictionary of all the targets and then lean on that. 

Would like to hear if this is an improvement or a step backward.

### Test Coverage


muchzill4:
If only there weren't duplicates in `product_code`. ;)

rosiemiller2:
> If only there weren't duplicates in `product_code`. ;)

Ah maybe this is where I have not kept up to date. I thought that there would only ever be a one to one mapping of `MbedTarget` to product code.

muchzill4:
@rosiemiller2 I mean this is a totally unhealthy state of things, but unfortunately sad reality last time I checked. I think modules have duplicate entries with boards or something like that. Worth double checking, as I'm not 100% positive that is still the case.

rosiemiller2:
Rob and I had a chat with Graham the other day and I think that there are multiple potential 'build_targets' (ie. the target name that is given to the build command) based on a combination of board types and build variants, but as far as I understand they will all be returned as the same `MbedTarget` based on their having the same product code.

mbed_targets/mbed_targets.py
```diff
@@ -57,15 +57,11 @@ def __init__(self) -> None:
         Raises:
             TargetDatabaseError: Failed to get the target data.
         """
-        self._target_data: List[dict] = target_database.get_target_data()
+        target_data = target_database.get_target_data()
+        self._all_targets = {target.product_code: target for target in [MbedTarget(target) for target in target_data]}
```
mark-edgeworth: Wow, that's quite a statement right there. Can this be simplified?

rwalton-arm: Might be clearer as a plain `for` loop rather than a nested comprehension.

madchutney:
@rosiemiller2 I added in a missing unit test to bring the coverage up (not your change but something omitted earlier) and get a green light. I hope you don't mind.

rwalton-arm:
I've thought about this a bit and I believe we should leave the code how it is. We've lost a lot of flexibility removing the `__iter__` method (for example this interface could've been used for something like producing the target validation report and now that's not possible). The current code behaves like a standard container (try to use `next` directly on a list or dictionary for example). Conceptually it makes sense as an iterable as this object is basically a list of targets. I don't see the benefit of removing the iterable. We could keep the dictionary stuff in the constructor for efficiency, but that also feels a bit like premature optimization when our mantra is supposedly 'maintainable code over performant code'. We have made this code more complex with the nested dictionary comp in the constructor, when currently have a simple, standard iterable class.

rosiemiller2:
@rwalton-arm @madchutney I am happy either way. I agree with Rob that the class makes sense as an iterable since it seems very intuitive to do`for mbed_target in mbed_targets`. I guess my thinking behind this change is that accessing an item from a dict is significantly cheaper than iterating through the list each time, and each time creating a new MbedTarget for each item processed. But this could be a premature optimisation while sacrificing other useful behaviour.

rosiemiller2:
I am tempted to shelve this for now and we can revisit it once we have more idea of how this class will be used.

madchutney:
I suggest that we add a generator function (not `__iter__`) which retains this functionality as it is true that exposing some kind of list all method will be useful in the future.

PR #51, opened 2020-02-06
📝 Add documentation for the auth_token

rwalton-arm:
### Description

Add package level docstring describing how to set the authorisation token.

### Test Coverage


mbed_targets/__init__.py
```diff
@@ -1,4 +1,24 @@
-"""Exposes the primary interfaces for the library."""
+"""mbed-targets provides an abstraction layer for boards and platforms supported by Mbed OS.
 
+This package is intended for use by developers of Mbed OS.
+
+To fetch data about private targets, an authentication token must be provided in an
```
rosiemiller2: The user will also need to have an account on os.mbed.com and be member of a vendor team that has permissions to see the private board.

mbed_targets/__init__.py
```diff
@@ -1,4 +1,24 @@
...
+environment variable named `MBED_API_AUTH_TOKEN`. The package also accepts a `.env` file
```
rosiemiller2: I know we tested this and I can't remember what the answer was, which one trumps if both are present?

rwalton-arm: The `.env` file wins as we use `override_system=true` - I'll add clarification on that

mbed_targets/__init__.py
rwalton-arm: Good point. I'll add a note about that.

mbed_targets/__init__.py
```diff
@@ -1,4 +1,29 @@
-"""Exposes the primary interfaces for the library."""
+"""mbed-targets provides an abstraction layer for boards and platforms supported by Mbed OS.
 
+This package is intended for use by developers of Mbed OS.
```
rosiemiller2: Sorry one more tiny thing:
```suggestion
This package is intended for use by developers using Mbed OS.
```

madchutney:
Looks good to me.

PR #72, opened 2020-02-19
Tweak sync_target_database script and add tests.

rwalton-arm:
### Description

Update the sync_target_database script to always push to the same git
branch. Only create a git commit when the database content changes.
Write a news item with the updated boards.

### Test Coverage


ci_scripts/test_sync_target_db.py
```diff
@@ -0,0 +1,270 @@
...
+    def test_news_file_item_text_formatting(self):
+        with mock.patch("ci_scripts.sync_target_database.Path", autospec=True) as mock_path:
+            mock_path().exists.return_value = False
+            text = sync_target_database.create_new_boards_news_item_text(json.loads(ONLINE_TARGET_DATABASE))
+            self.assertEqual(
+                text, "New boards added: MTB_UBLOX_NINA_B1, MTB_MTS_XDOT", "Text should be formatted correctly."
+            )
+
+    def test_news_file_name_suffix_increments_when_path_exists(self):
+        with mock.patch("ci_scripts.sync_target_database.Path", autospec=True) as mock_path:
```
muchzill4: Wonder if using something like `pyfakefs` wouldn't make this less implementation dependent (fragile).

ci_scripts/test_sync_target_db.py
```diff
@@ -0,0 +1,270 @@
+import json
+import os
+
+from collections import namedtuple
+from unittest import TestCase, mock
+
+from ci_scripts import sync_target_database
+
+ONLINE_TARGET_DATABASE = """[
```
muchzill4: Some gradual improvements I can see:
1. Have each targets json as variable, i.e: `target_1`, `target_2` and then do `ONLINE_TARGET_DATABASE=[target_1, target_2]` and `OFFLINE_TARGET_DATABASE=[target_1]`, etc.
2. Construct the `*_DATABASE` in each test, by using `target_1`, `target_2` json chunks.
3. _Maybe_ have target factory of sorts for tests using `factory_boy` or similar <- this is a stretch and not sure how useful since this is only used here.

rwalton-arm:
> Am I right in thinking this json dance is temporary as we're going to have offline mode that will use real target objects we could `__eq__`? If so, feel free to ignore my comments about that specific thing.
> 
> Suggestion to use `pyfakefs` is still valid I think.

Yeah, the json dance is temporary until the interface for offline mode is defined and implemented.  I'm considering implementing some variant of set operations for `MbedTargets` so we can take the difference between two target databases easily. I did look into `pyfakefs` earlier and it seemed a bit too complex for what I needed for the test cases. I will take another look and see if it's useful.

ci_scripts/test_sync_target_db.py
rwalton-arm: I will implement 1 and 2. I think 3 is too much extra complexity for now, and, as you mentioned, probably not that useful at the moment.

rwalton-arm: Refactored in 90fbe8253d3f23a6828c8e2121120f85b858bab7 - let me know what you think.

muchzill4: That looks better, `*_DATABASE` are still created outside of where they'd be used, which is what I was talking about in 2.

How about:
```python
TARGET_1 = {
}  # Regular dict
TARGET_2 = {
}  # Regular dict
```

And then:

```python
mock_online_targets.json_dump.return_value = json.dumps([TARGET_1, TARGET_2])
diffs = sync_target_database.diff_objects_from_json_str(
    mock_online_targets.json_dump(), json.dumps([TARGET_1])
)
self.assertEqual(diffs, [TARGET_2])
```

Thoughts?

rwalton-arm: I think that's a good suggestion. However, given the simplicity of the tests I think I'll leave it for now. I will make further improvements after I've got the offline mode implementation complete. Does that sound reasonable?

muchzill4: It does indeed.

ci_scripts/sync_target_database.py
```diff
@@ -45,31 +50,55 @@ def save_target_database(target_database_text: str) -> Path:
...
+def create_new_boards_news_item_text(new_targets: dict) -> str:
+    """Create a new news item string from the list of new targets."""
+    item_str = "New boards added:"
+    board_names = ", ".join(target["attributes"]["board_type"] for target in new_targets)
+    return f"{item_str} {board_names}"
+
+
+def write_news_file(item_text: str, file_name: str) -> Path:
+    """Write some text to a news file in the news directory of the repository."""
+    news_file_path = Path(configuration.get_value(ConfigurationVariable.PROJECT_ROOT), "news", f"{file_name}.misc")
```
acabarbaye:
```suggestion
    news_file_path = Path(configuration.get_value(ConfigurationVariable.PROJECT_ROOT), "news", f"{file_name}.feature")
```
This is to actually change the version of the package and have a record in the changelog. misc is only mentioned and does not change the version number

ci_scripts/sync_target_database.py
```diff
@@ -45,31 +50,55 @@ def save_target_database(target_database_text: str) -> Path:
...
+    inc = 0
+    while news_file_path.exists():
+        inc += 1
+        news_file_path = news_file_path.with_name(f"{file_name}{inc:0=2}.misc")
```
acabarbaye:
```suggestion
        news_file_path = news_file_path.with_name(f"{file_name}{inc:0=2}.feature")
```

ci_scripts/test_sync_target_db.py
```diff
@@ -0,0 +1,159 @@
...
+
+            def side_effect(arg):
+                mock_path().name = arg
+                return mock_path()
+
+            mock_path().with_name.side_effect = side_effect
+            mock_path().exists.side_effect = [True, True, False]
+            news_file_path = sync_target_database.write_news_file("news item", "news")
+            self.assertEqual(
+                "news02.misc", news_file_path.name, "News file name should have a incremented suffix of 02"
```
acabarbaye:
```suggestion
                "news02.feature", news_file_path.name, "News file name should have a incremented suffix of 02"
```

PR #76, opened 2020-02-24
✨ Support auto mode and expose get_target interface.

rwalton-arm:
### Description

Expose a `get_target` module level function which looks up a target by its
product code from either the online or offline database.

Support set difference operator in `MbedTargets` for easier
diffing of target databases. Using a non-symmetric difference should
allow users to easily detect if targets have been added to or deleted from
the local database when comparing with the online database.

Factor `product_code`, `board_type` and `platform_name` into `MbedTarget`
equality and hash methods. There is some duplication of product codes in
the database, so we would implicitly lose data when performing set operations
on the target database if we only take `product_code` into account.

At the moment the `get_target` function will return the first target with matching
product code it finds. There is a question around how we want to handle the 
product code duplication in the database here, or if we should fix this in the 
database itself.

### Test Coverage


mbed_targets/mbed_targets.py
```diff
@@ -84,22 +88,71 @@ class DatabaseMode(Enum):
 
     OFFLINE = 0
     ONLINE = 1
+    AUTO = 2
+
+
+def get_target(product_code: str, mode: DatabaseMode) -> MbedTarget:
```
muchzill4: Is `product_code` always `hex`? I wonder if we should take that into account in the representation of `MbedTargets` and when searching for target by it's `product_code`. I.e.: if we're just comparing strings, a `hex` product code of `0x123` which is equivalent to `0x0123` and `0123` won't match. While I'm skeptic about polluting `mbed-targets` with that knowledge, I'm wondering where the fact that product codes are of certain format should live and be documented.

mbed_targets/mbed_targets.py
```diff
@@ -84,22 +88,71 @@ class DatabaseMode(Enum):
...
+    """Base Interface to the Target Database."""
+
+    def __init__(self, target_data: Iterable) -> None:
+        """Initialise with a list of targets.
 
-    _target_data: Any
+        Args:
+            target_data: iterable of target data from a target database source.
+        """
+        self._target_data = list(MbedTarget(t) for t in target_data)
```
muchzill4: I wonder it if wouldn't be better to make `MbedTargets` inherit from `collections.abc.Set` (it already implements the interface anyway), since it seems to heavily rely on `set`- like operations. As I understand it:
- you'd get `__sub__` for free
- it'd make `_from_iterable` obsolete, as `set([1]) == set(set([1]))`

tests/test_mbed_targets.py
```diff
@@ -67,55 +67,69 @@ def test_cannot_compare_with_non_mbed_target_instance(self):
         target = _make_mbed_target(product_code="1000")
         self.assertFalse(target == "1000")
 
-    def test_hash_is_equal_to_hash_of_product_code(self):
-        tgt = _make_mbed_target(product_code="0100")
-        self.assertEqual(tgt.__hash__(), hash(tgt.product_code))
+    def test_hash_is_equal_to_hash_of_target_properties(self):
+        tgt = _make_mbed_target(product_code="0100", platform_name="a", board_type="b")
+        self.assertEqual(tgt.__hash__(), hash(tgt.product_code) ^ hash(tgt.platform_name) ^ hash(tgt.board_type))
```
muchzill4: Is `tgt.__hash__()` equivalent to `hash(tgt)`? Might be better to reduce the dunders. :)

tests/test_mbed_targets.py
```diff
@@ -67,55 +67,69 @@ def test_cannot_compare_with_non_mbed_target_instance(self):
...
-        tgt_2 = _make_mbed_target(product_code="0100", board_type="b")
+        tgt_2 = _make_mbed_target(product_code="0100", board_type="a")
+        self.assertEqual(tgt_1, tgt_2)
         tgts = dict()
         tgts[tgt_1] = "test"
         self.assertEqual(tgts[tgt_2], "test")
 
+    def test_repr_string_is_correctly_formed(self):
+        tgt = _make_mbed_target(board_type="a", product_code="b", platform_name="c")
+        self.assertEqual(tgt.__repr__(), f"{tgt.__class__.__name__}(board_type=a, product_code=b, name=c)")
```
muchzill4: Is `tgt.__repr__()` equivalent to `repr(tgt)`? Maybe reduce the need for dunders.
Also, I wonder if swapping `tgt.__class__.__name__` with `MbedTarget` makes sense here. 🤔

mbed_targets/mbed_targets.py
```diff
@@ -84,22 +88,71 @@ class DatabaseMode(Enum):
...
+
+
+def get_target(product_code: str, mode: DatabaseMode) -> MbedTarget:
+    """Get an MbedTarget by its product code.
+
+    Args:
+        product_code: the product code to look up in the database.
+        mode: a DatabaseMode enum field.
+    """
+    if mode == DatabaseMode.OFFLINE:
```
muchzill4: We can either make CodeClimate ignore this or extract 3 private functions. I guess this is a bit on the long side, but simple enough that I'm indifferent.

rosiemiller2:
I have nothing additional to add other than its looking really good! 👍

mbed_targets/mbed_targets.py
rwalton-arm: I've refactored it slightly in the next commit. If CodeClimate still has problems with the complexity we can tell it to ignore this.

PR #81, opened 2020-02-24
💚️ Use new targets interface in CI

rwalton-arm:
### Description

Update the sync target database CI scripts to use the new `MbedTargets` interface.

### Test Coverage


ci_scripts/sync_target_database.py
```diff
@@ -80,6 +78,20 @@ def write_news_file(item_text: str, file_name: str) -> Path:
...
+    """Creates and writes a news file from the added and removed targets.
+
+    Args:
+        added: added MbedTargets
+        removed: removed MbedTargets
+        news_file_name: name of the news file
+    """
+    news_item_text_added = create_news_item_text("Targets added: ", added)
+    news_item_text_removed = create_news_item_text("Targets removed: ", removed)
+    news_item_text = f"{news_item_text_added} {news_item_text_removed}"
```
muchzill4: Should there be some sort of separator between two messages? I'm not sure what'd work best. As it currently stands this would read: 
`New boards added: u-blox NINA-B1, Multitech xDOT New boards removed: NiceBoard1, HelloWorld-X3` - I can see it being hard to read the changelog.
How about: `Offline database update. Removed: u-blox NINA-B1, Multitech xDOT. Added: NiceBoard1, HelloWorld-X3`.

ci_scripts/sync_target_database.py
```diff
@@ -54,18 +53,17 @@ def save_target_database(target_database_text: str, output_file_path: Path) -> N
...
 
 
-def diff_objects_from_json_str(incoming_text: str, current_text: str) -> List[dict]:
-    """Convert two json strings to objects and take the difference."""
-    current = json.loads(current_text)
-    incoming = json.loads(incoming_text)
-    return [i for i in current + incoming if i not in current or i not in incoming]
+def get_boards_added_or_removed(offline_targets: MbedTargetsOffline, online_targets: MbedTargetsOnline) -> List[dict]:
+    """Check boards added and removed in relation to the offline target database."""
+    added = online_targets - offline_targets
```
muchzill4: Am I right in thinking that online db update to any of the properties `__eq__` relies on will trigger a removal and addition for the same board here?
This also reminds me of the fact that we tend to bump "mbed os support" features in bulk when new version of it is released. Food for thought if this should be catered for - it'd probably yield updates to hundreds of targets.

ci_scripts/sync_target_database.py
```diff
@@ -140,18 +152,20 @@ def main(args: argparse.Namespace) -> int:
...
     try:
-        targets = MbedTargets().json_dump()
+        online_targets = MbedTargetsOnline()
+        news_file_name = f"{datetime.date.today().strftime('%Y%m%d')}"
         if TARGET_DATABASE_PATH.exists():
-            diff_objects = diff_objects_from_json_str(targets, TARGET_DATABASE_PATH.read_text())
-            if not diff_objects:
+            offline_targets = MbedTargetsOffline()
+            added, removed = get_boards_added_or_removed(offline_targets, online_targets)
+            if not len(added) and not len(removed):
```
muchzill4:
```suggestion
            if not (added or removed)
```

No need for `len()`.

ci_scripts/test_sync_target_db.py
```diff
@@ -1,147 +1,123 @@
...
-        with mock.patch("ci_scripts.sync_target_database.MbedTargets") as mock_online_targets:
-            mock_online_targets.json_dump.return_value = ONLINE_TARGET_DATABASE
-            diffs = sync_target_database.diff_objects_from_json_str(
-                mock_online_targets.json_dump(), OFFLINE_TARGET_DATABASE
-            )
-            self.assertEqual(diffs, json.loads(EXPECTED_DIFF), "Actual diff and expected diff should match")
+    def test_get_boards_added_or_removed_detects_added(self):
+        mock_online_targets, mock_offline_targets = _make_mbed_targets_for_diff([TARGET_1, TARGET_2], [TARGET_1])
+        added, removed = sync_target_database.get_boards_added_or_removed(mock_offline_targets, mock_online_targets)
+        self.assertEqual(len(added), 1, "One new board to be added to offline db.")
```
muchzill4: NIT: I'm a bit conflicted about those custom error messages. I.e.:
1. `One new board to be added to offline db.` -> `Expected exactly one new board to be added to offline db.`
2. This one in particular: `No boards removed from offline db.` -> `Expected no boards to be removed from offline db.`
3. `Should return an empty targets container when no targets added` -> `Returned a non-empty target container when no targets added`.
Etc.

NIT^2: "Should" sounds like asking, yet in my mind computer must obey what I tell it to do. It's not a _maybe_, it's an _expectation_. Not only "should" takes up space, but is meaningless in the grand scheme of things. "Should return" vs "returns"; "should to something" vs "does something" etc. 😉

.codeclimate.yml
```diff
@@ -33,3 +33,5 @@ checks:
 exclude_patterns:
   - "ci_scripts/target_validation_report.py"
   - "tests/test_mbed_targets.py"
+  - "ci_scripts/sync_target_database.py"
```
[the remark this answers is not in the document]
rwalton-arm: I listed them here because I think the ci scripts shouldn't be under the same level of autobot scrutiny as our actual production code. I don't know how to turn off checks in the CC UI.

ci_scripts/sync_target_database.py
rwalton-arm: I'll add a separator. I'm not sure using the term "boards" is the best thing here so I think I'll stick with the current wording, with separator.

ci_scripts/sync_target_database.py
rwalton-arm: I think changing the '__eq__' method to check for mbed os support is outside of the scope of this PR. It may be the right thing to do.

ci_scripts/sync_target_database.py
muchzill4: Oh, I was trying to mimic what the current message would be and failed. 😢 
For posterity: 
`Targets added:  u-blox NINA-B1, Multitech xDOT Targets removed: NiceBoard1, HelloWorld-X3` needs a separator as it's not readable.

ci_scripts/test_sync_target_db.py
rwalton-arm: I will remove "should" - although I'm not convinced adding the word 'expected' to the other messages improves them. This is the point where I say we should be using pytest anyway but I'm not going to go there :)

ci_scripts/sync_target_database.py
rwalton-arm: This has a syntax error, you missed the colon. Fixed and applied anyway :)

muchzill4: I realized that too, sorry. 😅

ci_scripts/sync_target_database.py
rwalton-arm: Added a full stop in 793e31c

.codeclimate.yml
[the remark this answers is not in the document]
muchzill4: I'm not sure what's best here. The fact we can't do per-line ignores that are respected by the cli bothers me slightly. Ignoring entire files bothers me more.

I can show you how to turn off checks in CC UI.

ci_scripts/test_sync_target_db.py
```diff
@@ -1,148 +1,122 @@
...
-        with mock.patch("ci_scripts.sync_target_database.MbedTargets") as mock_online_targets:
-            mock_online_targets.json_dump.return_value = ONLINE_TARGET_DATABASE
-            diffs = sync_target_database.diff_objects_from_json_str(
-                mock_online_targets.json_dump(), OFFLINE_TARGET_DATABASE
-            )
-            self.assertEqual(diffs, json.loads(EXPECTED_DIFF), "Actual diff and expected diff should match")
+    def test_get_boards_added_or_removed_detects_added(self):
+        mock_online_targets, mock_offline_targets = _make_mbed_targets_for_diff([TARGET_1, TARGET_2], [TARGET_1])
+        added, removed = sync_target_database.get_boards_added_or_removed(mock_offline_targets, mock_online_targets)
+        self.assertEqual(len(added), 1, "Expect one new board to be added to offline db.")
```
muchzill4: If only we were able to do: `added == MbedTargets([TARGET_1])`. 🤔

PR #83, opened 2020-02-28
✨ Initial loading of targets.json file

rosiemiller2:
Added code for loading in targets.json from given file location and doing a basic lookup in the resulting dictionary.

I have also changed name of the targets.json data file (ie the local db file) so as not to confuse with targets.json being pulled in from mbed-os.

Next step will be to properly parse the targets.json file using the inheritance as what is returned now is just the 'bottom level' in the hierarchy. 

### Test Coverage


mbed_targets/_internal/target_database.py
```diff
@@ -18,7 +18,7 @@
 
 def get_target_database_path() -> pathlib.Path:
     """Return the path to the offline target database."""
-    return pathlib.Path(INTERNAL_PACKAGE_DIR, "data", "targets.json")
+    return pathlib.Path(INTERNAL_PACKAGE_DIR, "data", "offline_target_list.json")
```
rwalton-arm: We'll also need to change the path in `sync_target_database` as it uses a path derived from `ConfigurationVariable.PROJECT_ROOT`... or we should fix `sync_target_database` so it uses this function.

mbed_targets/_internal/target_attributes.py
```diff
@@ -0,0 +1,82 @@
...
+    Raises:
+        TargetAttributesNotFoundError: there is no target attribute data found for that target.
+    """
+    try:
+        return all_targets_data[target_name]
+    except KeyError:
+        raise TargetAttributesNotFoundError(f"Target attributes for {target_name} not found.")
+
+
+def get_target_attributes(path_to_targets_json: str, target_name: str) -> Any:
```
rwalton-arm: In the past I've usually tried to follow a convention (where possible) where the "public" methods or free functions are placed at the top of the file. This is so it's easier to see the public API without trawling through the "private" helpers. I think it would be good to move this function up to the top of the file below the exception classes.

mbed_targets/_internal/target_database.py
rosiemiller2: Ah cool, yes I'll take a look at that. I also liked your description in the last PR of that file being an 'offline database snapshot' so I might tweak the name too. 👍

PR #84, opened 2020-02-28
Make MbedTargets inherit from Set.

rwalton-arm:
### Description

Inheriting from `Set` means we get all set operations and rich 
comparison operators for free.

The initialiser has been changed slightly to allow us to inherit from
`Set`, it now takes an `Iterable[MbedTarget]` which is created in the
subclasses `__init__` and passed up to the parent.

### Test Coverage


mbed_targets/mbed_targets.py
```diff
@@ -124,22 +124,27 @@ class UnsupportedMode(ToolsError):
...
 
-    def __init__(self, target_data: Iterable[dict]) -> None:
+    Don't instantiate an MbedTargets directly, this class is intended for subclassing.
+
+    Subclasses are expected to override __init__ to create an iterable of MbedTarget objects and pass it to
+    the parent initialiser via super().__init__.
+    """
+
+    @classmethod
+    def _from_iterable(cls: object, it: Iterable) -> "MbedTargets":
```
rosiemiller2: Nit: I personally prefer `iterable` to `it` since it makes parsing very easy. Though I don't know if `it` is an established convention here?

rwalton-arm: Not really a convention, just how it's defined in the `Set` base class. Might be worth changing it to `iterable` though.

mbed_targets/mbed_targets.py
```diff
@@ -124,22 +124,27 @@ class UnsupportedMode(ToolsError):
...
+        """Override Set._from_iterable to always return an MbedTargets instance and not a subclass instance."""
+        return MbedTargets(it)
```
rosiemiller2: So this is needed because the constructors in the subclasses don't take an iterable. I guess the gotcha here is that if you add additional methods/attributes to the subclass the result of say a union action wouldn't contain instances of the subclass but of MbedTargets which wouldn't have these methods. 
Might be worth making this super clear so people coming later on don't do this accidentally.

PR #90, opened 2020-03-05
🐛 Use correct attribute for type

muchzill4:
### Description

Silly me for not checking what `board_type` really means. Turns out `target_type` is the source of truth for `module|platform` value. ;)

### Test Coverage


rosiemiller2:
Just to check I understand - `board_type = eg. k64f` and `target_type = platform/module/component`?

muchzill4:
> Just to check I understand - `board_type = eg. k64f` and `target_type = platform/module/component`?

That is correct. Maybe we should call it something else? 🤔

rosiemiller2:
It is a tad confusing. What do you think about `target_group` or `target_category`? They aren't much better but are different enough from `board_type` to help.

muchzill4:
@rosiemiller2 Are you ok with me merging this as is? 

If anything, I think `board_type` is a misnomer. At the moment, it is used for variants only and the offline discussion we've had about potentially exposing `variants` in the api would make it obsolete.
Another issue is, we seem to follow/expose the api field names semi-rigorously (looking at you `board_name`). I'm a bit scared of making a naming decision right now and/or change _unrelated_ names.

rosiemiller2:
Yep I'm happy with it going in the way it is 👍

PR #91, opened 2020-03-05
✨ Parse attributes

rosiemiller2:
### Description

Add in the parsing functionality for parsing the targets.json hierarchy.

To see the original parsing code see https://github.com/ARMmbed/mbed-os/blob/master/tools/targets/__init__.py. 

To see the documented rules see https://os.mbed.com/docs/mbed-os/latest/reference/adding-and-configuring-targets.html.

This isn't all hooked up together yet so the complete flow isn't ready but hopefully you can get the idea. 

### Test Coverage


mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Returns:
+        The accumulator object with the new elements added
+    """
+    for element in elements_to_add:
+        accumulator[attribute_name].append(element)
+    return accumulator
+
+
+def _remove_attribute_element(
```
rosiemiller2: This and `_add_attribute_element` look very similar at this stage but `_remove_attribute_element` will have to handle more complexity and so will change quite dramatically in the next PR.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Returns:
+        A list of dicts representing each target in the hierarchy.
+    """
+    accumulation_order: List[dict] = []
+
+    still_to_visit: Deque[dict] = deque()
+    still_to_visit.appendleft(all_targets_data[target_name])
+
+    while still_to_visit:
```
rosiemiller2: This and the hierarchy calculator for overriding attributes always seem to end up being pretty difficult to understand/read. I tried pulling bits out into separate functions but that seemed to make it even worse, though now Code Climate is unhappy about the complexity. Any suggestions?

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Args:
+        accumulation_order: the order in which the targets inherit from the target itself to its highest ancestor
+
+    Returns:
+        A dictionary containing all the accumulating attributes for a target
+    """
+    accumulated_attributes = {}
+    for attribute_name in ACCUMULATING_ATTRIBUTES:
+        for target in accumulation_order:
```
rwalton-arm: We could look into using `itertools.product` here instead of a nested `for`.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rwalton-arm: If there's any possibility of a recursive solution here that would probably be easier to grok. I'm still trying to figure out exactly what's going on here in my rusty brain so I don't have any concrete suggestions yet.

rosiemiller2: There is at least for the overriding attributes hierarchy parser, but I have a bit of a hatred for recursive solutions, though I do understand that is probably just me. I could have a go at that and see if it ends up a bit easier to understand.

mbed_targets/_internal/target_attribute_hierarchy_parsers/overriding_attribute_parser.py
```diff
@@ -0,0 +1,95 @@
...
+
+    Returns:
+        A list of dicts representing each target in the hierarchy.
+    """
+    override_order: List[dict] = []
+
+    still_to_visit: Deque[dict] = deque()
+    still_to_visit.appendleft(all_targets_data[target_name])
+
+    while still_to_visit:
```
rwalton-arm: We should factor out the common code from here and `_targets_accumulate_hierarchy` into a more generic generator function.

rosiemiller2: I thought that might make it even more impenetrable 😂

mbed_targets/_internal/target_attribute_hierarchy_parsers/overriding_attribute_parser.py
```diff
@@ -0,0 +1,95 @@
...
+
+    Returns:
+        A dictionary containing all the overridden attributes for a target
+    """
+    target_attributes = reduce(lambda x, y: _update_dict(x, y), reversed(override_order))
+    overridden_attributes = _remove_accumulating_attributes(target_attributes)
+    return overridden_attributes
+
+
+def _update_dict(dict_1: Dict[Any, Any], dict_2: Dict[Any, Any]) -> Dict[Any, Any]:
```
rwalton-arm: I'd prefer this was called `merge_dicts` or something as the name `update_dict` makes me think we're updating the same dictionary object rather than returning a new dict.

mbed_targets/_internal/target_attribute_hierarchy_parsers/overriding_attribute_parser.py
```diff
@@ -0,0 +1,95 @@
...
+    """Helper function to return an updated dictionary without modifying the original two.
+
+    Args:
+        dict_1: dictionary to be updated (will be overridden by dict_2 if there are clashes)
+        dict_2: dictionary containing new updates
+
+    Returns:
+        The equivalent of dict_1 if we had run dict_1.update(dict_2).
+    """
+    temp_dict = dict_1.copy()
```
rwalton-arm: I think we should change the name from `temp_dict` to `output_dict` as this is returned so it isn't temporary.

mbed_targets/_internal/target_attribute_hierarchy_parsers/overriding_attribute_parser.py
```diff
@@ -0,0 +1,95 @@
...
+    temp_dict.update(dict_2)
+    return temp_dict
```
rwalton-arm: I think an even better solution might be to use the fancy dict comprehension syntax here, so something like `return {**a, **b}`

mbed_targets/_internal/target_attribute_hierarchy_parsers/overriding_attribute_parser.py
```diff
@@ -0,0 +1,95 @@
...
+def _remove_accumulating_attributes(target_attributes: Dict[str, Any]) -> Dict[str, Any]:
+    """Removes all the accumulating attributes and their modifiers.
+
+    Args:
+        target_attributes: a dictionary of attributes for a target
+
+    Returns:
+        The target attributes with the accumulating attributes removed.
+    """
+    temp_dict = target_attributes.copy()
```
rwalton-arm: Again I'd change the name of this to `output_dict` as it's not temporary.

mbed_targets/_internal/target_attribute_hierarchy_parsers/overriding_attribute_parser.py
rosiemiller2: Ooh I have never seen that before. I will take a look!

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rwalton-arm: Sometimes recursive solutions are terrible :) it just depends on the code, sometimes they can simplify code like this.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rosiemiller2: I've had a go at refactoring this. Seems like a bit of an improvement but not entirely sure.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rosiemiller2: Ok, I had a go at doing this recursively and I still ended up needing a queue in order to track previously traversed 'inherits' and it ended up more complicated than this (similar discussion here: https://cs.stackexchange.com/questions/64379/can-breadth-first-search-be-implemented-recursively-without-data-structures). The depth first search done in overriding attributes hierarchy can be done recursively, but part of me really likes the symmetry between the functions here.

mbed_targets/_internal/target_attribute_hierarchy_parsers/overriding_attribute_parser.py
rosiemiller2: I've had a think about this and I think we shouldn't try and factor out too much. This is a very self-contained algorithm, I don't want to have to be looking between files working out what is part of the shared code and what isn't. I also don't want the two parsers to be reliant on each other. I want to make it very easy in the future to change either/both of their functionality (I am not sure we want there to be two different ways of finding the hierarchy anyway).

rwalton-arm: I still think we should definitely factor out the repeated code here. It will make it easier to understand if it's wrapped in a nicely named generator and reused in these two functions. In some _rare_ cases it's ok to ignore DRY and repeat some code. I don't think this is one of those situations though.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rosiemiller2: Looked at it again and it definitely is an improvement - thanks for the suggestion 👍

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
```diff
@@ -0,0 +1,150 @@
...
+        accumulation_order: the order in which the targets inherit from the target itself to its highest ancestor
+
+    Returns:
+        A dictionary containing all the accumulating attributes for a target
+    """
+    accumulated_attributes = {}
+
+    target_attribute_pairs = itertools.product(accumulation_order, ACCUMULATING_ATTRIBUTES)
+    filtered_pairs = ((target, attribute) for target, attribute in target_attribute_pairs if attribute in target.keys())
+    for target, attribute_name in filtered_pairs:
```
rwalton-arm: I think this is a bit hard to read, could you not do something like:
```
for tgt, attr in itertools.product(accumulation_order, ACCUMULATING_ATTRIBUTES):
    if attr in tgt.keys():
        accumulated_sttrs.update(_calc...)
```
Or does that not work?

rosiemiller2: Yeah that is much better, thanks - Ill get there in the end 😂

rosiemiller2: 3beb9db

mbed_targets/_internal/target_attribute_hierarchy_parsers/overriding_attribute_parser.py
rosiemiller2: After IRL walk through of the code I think that factoring out will add more complexity. Fingers crossed this can be simplified the correct way in the near future - by only having one way of inheriting. 🤞

PR #94, opened 2020-03-06
♻ Convert MbedTarget to a dataclass

rwalton-arm:
### Description
`MbedTarget` was becoming hard to maintain due to having to update the `__eq__`, `__hash__` methods and tests every time we added a new field.

Refactor `MbedTargets` into a `dataclass` to remove the maintenance burden.

### Test Coverage


mbed_targets/mbed_targets.py
```diff
@@ -30,101 +30,46 @@
...
-        """Create a new instance of a target.
+    board_type: str
+    board_name: str
+    product_code: str
+    target_type: str
+    slug: str
+    build_variant: Tuple[Any, ...]
+    mbed_os_support: Tuple[Any, ...]
+    mbed_enabled: Tuple[Any, ...]
+    target_entry: dict = field(default_factory=dict, repr=False, compare=False)
```
muchzill4: What's the reason for having `target_entry` as one of the fields? Can it not be discarded after we've created the instance?

rwalton-arm: It's so we can serialise the target back to JSON. I thought about trying to manage the raw entries in `MbedTargets` but I think that would get messy as we'd have to change the initialiser and hack the `Set._from_iterable` classmethod.

mbed_targets/mbed_targets.py
```diff
@@ -30,101 +30,46 @@
...
     """Representation of an Mbed Target."""
 
-    def __init__(self, target_database_entry: dict):
-        """Create a new instance of a target.
+    board_type: str
+    board_name: str
+    product_code: str
+    target_type: str
+    slug: str
+    build_variant: Tuple[Any, ...]
```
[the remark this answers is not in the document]
rwalton-arm: mypy says no!

rosiemiller2:
I agree with the questions above but apart from that looks good to me 💯

mbed_targets/mbed_targets.py
rwalton-arm: If you have any ideas for decoupling the serialisation from this object do tell. I dislike this bit of code too.

mbed_targets/mbed_targets.py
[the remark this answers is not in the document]
muchzill4: `EXPLAIN;` 🤔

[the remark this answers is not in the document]
rosiemiller2: Mypy explains itself to no man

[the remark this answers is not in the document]
rwalton-arm: I did explain, it said no!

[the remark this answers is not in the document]
rwalton-arm: 68c70a8

mbed_targets/mbed_targets.py
muchzill4: How about `to_target_entry` which mirrors `from_target_entry`. I do realize that we'd lose some information when converting from from objects returned by the api (like `id`, etc), but since we don't use this information, do we really care? Obviously, changing how this behaves would require us to re-sync the db. :)

tests/test_mbed_targets.py
```diff
@@ -256,58 +202,40 @@ def test_get_target_failure(self, mocked_get_target_data):
...
-        tgts_off = MbedTargets.from_offline_database()
-        tgts_on = MbedTargets.from_online_database()
-        diff_base_on = tgts_on - tgts_off
-        diff_base_off = tgts_off - tgts_on
+        mocked_get_online_target_data.return_value = raw_target_data
 
-        self.assertEqual(len(diff_base_on), 1)
-        self.assertEqual(len(diff_base_off), 0)
+        targets = [MbedTarget.from_target_entry(t) for t in raw_target_data]
+        filtered_target_data = [target.__dict__ for target in targets]
```
muchzill4: Is `__dict__` equivalent to `asdict`? If so, I'd rather we avoid using dunders where possible.

tests/test_mbed_targets.py
```diff
@@ -256,58 +202,40 @@ def test_get_target_failure(self, mocked_get_target_data):
...
 
-        self.assertTrue(isinstance(tgts_one - tgts_two, MbedTargets))
-
-    def test_difference_operator_raises_for_non_targets_instance(self, mocked_offline_target_data):
-        mocked_offline_target_data.return_value = _make_dummy_internal_target_data()
-
-        tgts = MbedTargets.from_offline_database()
+        self.assertEqual(
+            json_str_from_raw,
+            json.dumps([t1_raw.__dict__, t2_raw.__dict__], indent=4),
```
muchzill4: Is `__dict__` equivalent to `asdict`? If so, I'd rather we avoid using dunders where possible.

tests/test_mbed_targets.py
```diff
@@ -256,58 +202,40 @@ def test_get_target_failure(self, mocked_get_target_data):
...
+            "JSON string should match serialised target __dict__.",
+        )
 
-        with self.assertRaises(TypeError):
-            tgts - 1
+        self.assertEqual(json_str_from_filtered, json.dumps([t1_filt.__dict__, t2_filt.__dict__], indent=4))
```
muchzill4: Is `__dict__` equivalent to `asdict`? If so, I'd rather we avoid using dunders where possible.

tests/test_mbed_targets.py
rwalton-arm: `asdict` just calls `object.__dict__` AFAIK. I'll use the `asdict` wrapper in the tests though.

PR #99, opened 2020-03-11
✨ Extract labels

rosiemiller2:
### Description

The current tools require a list of labels for each target which are the names of all the targets it inherits from. I am not entirely sure how this is used at the moment so adding this functionality to get going and will modify it later if different things are needed (eg. if the order is important I will need to do this differently).

To keep labels separate from the other attributes have created an MbedAttributes object to contain both. This will be renamed in the coming PRs to BuildTarget or something as discussed in yesterdays meeting.

### Test Coverage


mbed_targets/_internal/target_attributes.py
```diff
@@ -30,6 +32,17 @@ class TargetAttributesNotFoundError(TargetAttributesError):
     """Attributes for target not found in targets.json."""
 
 
+@dataclass
```
rwalton-arm: Nice!

rwalton-arm: Do we plan on modifying the attribute values later on? If not it might be worth setting `frozen=True` to make the object immutable.

mbed_targets/_internal/target_attributes.py
```diff
@@ -30,6 +32,17 @@ class TargetAttributesNotFoundError(TargetAttributesError):
...
+class MbedTargetAttributes:
+    """A set of build attributes for an Mbed Target.
+
+    As defined in the mbed-os repository's targets.json config file.
```
rwalton-arm: I'd prefer we had an `Attributes` section in this docstring describing what the attributes are. Even though this is a simple object it's still good practice to document the attributes.

rosiemiller2: ed78110

mbed_targets/_internal/target_attributes.py
rosiemiller2: We shouldn't ever be modifying it. I will ❄️ it!

rosiemiller2: ebc4dcf

PR #101, opened 2020-03-12
♻ Refactor exception hierarchy.

rwalton-arm:
### Description

This patch moves the exceptions to their own files in the public and
internal packages. This is so we can create a distinction between
exposed exceptions and internal exceptions. This should make the public
API more sensible in regards to exception handling.

### Test Coverage


mbed_targets/mbed_targets.py
```diff
@@ -2,14 +2,15 @@
...
-- `get_target_by_online_id` function, which looks up an MbedTarget by its slug and type
-The lookup can be from either the online or offline database, depending on the given mode.
-The mode can be one of the following DatabaseMode enum fields:
-    AUTO: the offline database is searched first, if the target isn't found the online database is searched.
-    ONLINE: the online database is always used.
-    OFFLINE: the offline database is always used.
+An MbedTarget can be retrieved by calling one of the public functions. All of which return an instance
+of MbedTarget.
+
+Functions:
```
muchzill4: This is unnecessary as pdoc will generate a list of public members of the module.

tests/test_mbed_targets.py
```diff
@@ -238,6 +241,16 @@ def test_json_dump_from_raw_and_fitered_data(self, mocked_get_offline_target_dat
 
         self.assertEqual(json_str_from_filtered, json.dumps([t1_filt.__dict__, t2_filt.__dict__], indent=4))
 
+    @mock.patch("mbed_targets._internal.target_database.get_offline_target_data")
+    def test_handles_database_errors(self, mocked_get_offline_target_data, mocked_get_online_target_data):
```
muchzill4: I think this should be separated into two separate tests (and potentially TestCases). Seems to me we're trying too hard to fit all tests under `TestMbedTargets` umbrella.

rwalton-arm: Do you mind if that's done in another PR? After this and the CI fix are in I would like to rationalise the tests as I think they're nonsensical in some cases.

mbed_targets/mbed_targets.py
```diff
@@ -126,13 +120,27 @@ class MbedTargets(Set):
...
     def from_online_database(cls) -> "MbedTargets":
-        """Initialise with the online target database."""
-        return cls(MbedTarget.from_target_entry(t) for t in target_database.get_online_target_data())
+        """Initialise with the online target database.
+
+        Raises:
+            TargetDataNotFound: Could not retrieve data from the target database.
+        """
+        try:
+            return cls(MbedTarget.from_target_entry(t) for t in target_database.get_online_target_data())
```
muchzill4: FWIW, I think we should wrap only the code raising exception in `try/except` here. It's hard to figure out what can raise `TargetDatabaseError` by just reading this.

mbed_targets/mbed_targets.py
madchutney: I'd be worried that this won't be maintained and will get out of date.

tests/test_mbed_targets.py
muchzill4: Aye.

mbed_targets/mbed_targets.py
```diff
@@ -126,13 +120,27 @@ class MbedTargets(Set):
...
-        """Initialise with the offline target database."""
-        return cls(MbedTarget(**t) for t in target_database.get_offline_target_data())
+        """Initialise with the offline target database.
+
+        Raises:
+            TargetDataNotFound: Could not retrieve data from the target database.
+        """
+        try:
+            return cls(MbedTarget(**t) for t in target_database.get_offline_target_data())
+        except TargetDatabaseError as tgt_db_err:
```
madchutney: I don't really understand why we need to change the exception,

rwalton-arm: I didn't want to expose internal exceptions to the user. I was discussing this on Slack with @muchzill4 and he asked similar questions, so maybe I should make the `TargetDatabaseError` public? Although all the internal exceptions derive from it so I thought that was weird, but I guess it's the same for `ToolsError`.

mbed_targets/mbed_targets.py
rwalton-arm: Agreed, deleting this as pdoc should auto generate something useful anyway.

mbed_targets/mbed_targets.py
rwalton-arm: 64441d3

mbed_targets/mbed_targets.py
rwalton-arm: We don't transform the exception anymore so this `try/except` has been removed.

mbed_targets/mbed_targets.py
rwalton-arm: I merged the master docstring changes and removed the functions/classes lists. We still list the entry points. Thoughts?

muchzill4: If those entry points are public functions listed by pdoc it's not necessary to leave them there. It'll become an unnecessary maintenance burden.

rwalton-arm: Yes, I was just looking at the pdoc artifacts and it makes no sense to list them. Removing in the next commit.

rwalton-arm: ec5e0a0

mbed_targets/mbed_targets.py
```diff
@@ -123,12 +121,20 @@ class MbedTargets(Set):
 
     @classmethod
     def from_offline_database(cls) -> "MbedTargets":
-        """Initialise with the offline target database."""
+        """Initialise with the offline target database.
+
+        Raises:
+            TargetDatabaseError: Could not retrieve data from the target database.
```
muchzill4: When does it raise this error?

rwalton-arm: Raised in the call to `get_*_database` although I think it's relevant to the API as users are exposed to it, so it should have some documentation.

muchzill4: I'm following now. 👍

PR #103, opened 2020-03-12
Fix sync_target_database

rwalton-arm:
### Description

Now we use tuples for the MbedTarget fields, all targets were comparing != after
decoding the local database. This is because tuples are serialised to JSON arrays
which are deserialised back to lists.

Add a custom decode function which converts the lists to tuples.

### Test Coverage


muchzill4:
I'm a bit confused why this is required for offline but not online mode. Both use json as input, I'd think the fact we use tuples should be handled by `MbedTargets` and/or `MbedTarget` abstraction layer.

muchzill4:
Yeah, having looked at it I think converting to tuples should not happen here. 
- There's `MbedTargets.json_dump` which converts target database to json
- There's `MbedTarget.from_target_entry` which creates target from online json
- With this PR, there's details of how to de-serialize `json_dump`ed data back to valid `MbedTarget` elsewhere

I think it'd be nice to have `MbedTarget.to_json` and `MbedTarget.from_json` which would know how to serialize/de-serialize target data into/from offline backup.

rwalton-arm:
> I like leaving random comments and requesting changes afterwards, because logic is unnecessary. 😅

Yes, who needs logic. Anyway I'll add another classmethod to `MbedTarget` which does this tuple conversion. I don't want to add a `to_json` method to `MbedTarget` as I don't think an `MbedTarget` should care about reformatting for the database, that should be the responsibility of the `MbedTargets` class as that essentially manages the database. The target is just supposed to be a dumb data container that can be constructed from various data sources.

PR #105, opened 2020-03-13
✨ Add a method to fetch build attributes on MbedTarget

rosiemiller2:
### Description

Add a method for fetching attributes onto the MbedTarget object. 
Add a couple of integration-style tests for this method.

### Test Coverage


mbed_targets/mbed_targets.py
```diff
@@ -90,6 +91,26 @@ def from_offline_target_entry(cls, target_entry: dict) -> "MbedTarget":
...
+        These attributes contains the specific information needed to build Mbed applications for this Mbed Target.
+
+        Args:
+            path_to_targets_json: path to a targets.json file found in the Mbed OS library
+
+        Returns:
+            A dict containing the parsed attributes from targets.json
+
+        Raises:
+            FileNotFoundError: path provided does not lead to targets.json
```
rwalton-arm: It looks like we suppress the `FileNotFoundError` and raise `TargetBuildAttributesError`, so `FileNotFoundError` isn't really relevant to the API here.

mbed_targets/mbed_targets.py
```diff
@@ -90,6 +91,26 @@ def from_offline_target_entry(cls, target_entry: dict) -> "MbedTarget":
...
+            TargetBuildAttributesError: an error has occurred while fetching build attributes
+        """
+        try:
+            return target_attributes.get_target_attributes(path_to_targets_json, self.board_name)
+        except (FileNotFoundError, target_attributes.TargetAttributesError) as e:
+            raise TargetBuildAttributesError(e)
```
rwalton-arm: Should we use `raise from` here so we preserve the full traceback?

mbed_targets/mbed_targets.py
rosiemiller2: 56383a8

mbed_targets/mbed_targets.py
rosiemiller2: 56383a8

tests/test_mbed_targets.py
```diff
@@ -11,9 +11,13 @@
     MbedTargets,
     _get_target,
     _target_matches_query,
+    TargetBuildAttributesError,
 )
 from mbed_targets._internal.configuration import DatabaseMode
 
+TEST_DIR = pathlib.Path(__file__).parents[0]
+PATH_TO_TEST_TARGETS_JSON = TEST_DIR.joinpath("test_mock_sources/test_targets.json")
```
muchzill4: Having a separate fixture file like this is a bit fragile. Could we not `pyfakefs` and pass the path to file generated in test instead?

tests/test_mbed_targets.py
```diff
@@ -85,6 +89,31 @@ def test_empty_database_entry(self):
         self.assertEqual("", mbed_target.target_type)
         self.assertEqual("", mbed_target.slug)
 
+    def test_fetch_target_build_attributes(self):
+        mbed_target = _make_mbed_target(board_name="Target_3")
+        expected_result = {
```
muchzill4: Another reason why external `subject` is not a good pattern. Ideally the test case should contain all the sample data you need to operate on.

tests/test_mbed_targets.py
rosiemiller2: Yeah that makes sense, Ill give that a go 👍

rwalton-arm:
I'm wondering if we should just have a `build_attributes` field in `MbedTarget` and then construct the target with the build attributes rather than calling a method on the target. I'd prefer the `MbedTargets` container dealt with the data fetching, while `MbedTarget` remained a passive data structure.

rosiemiller2:
Hmm, I know what you mean but my thought process is that I don't want to have to require a path to mbed os library's target.json every time we create an MbedTarget. Mbed devices won't need any of the data from targets.json for example, so why should it fail if I don't provide a path.  Also, if it fails for whatever reason, it will mix up the 'bad targets.json reading' with the object creation which I think would be bad UX.

rosiemiller2:
An alternative could be just an exposed function `get_build_attributes` which takes an MbedTarget and a path as an argument and returns a dict. That way MbedTargets remains purely data.

rwalton-arm:
> An alternative could be just an exposed function `get_build_attributes` which takes an MbedTarget and a path as an argument and returns a dict. That way MbedTarget remains purely data.

I like this alternative as it decouples the operations on the database/json from the data structure. I wonder if we should actually have a `TargetBuildAttributes` struct instead of returning a dict so we have easier discoverability, and a consistent interface, for the attributes? If we need to iterate over the dict keys and the keys aren't always the same, then we should return a dict. If the keys are fixed I'd prefer a struct/dataclass.

rosiemiller2:
Yeah, there is no reliable pattern to the attributes that can be found in targets.json 😞, even the set of accumulating attributes that we have to recognise are not guaranteed to be there. 
Cool I will make that change 👍

tests/test_mbed_targets.py
rosiemiller2: 715d1b0

tests/test_mbed_targets.py
rosiemiller2: 715d1b0

PR #109, opened 2020-03-17
✨ Expose config variables

muchzill4:
### Description

@rwalton-arm pointed out that pdoc supports documenting variables. I must say I wasn't a fan of this approach initially, but it seems to flow quite nicely in the end. We don't need to introduce our wrappers and it reduces duplication quite severely.

Screenshot 2020-03-17 at 13 34 11

The exported `config_variables` will be proxied via `mbed-devices` to `mbed-tools` and displayed in the `cli`.

### Test Coverage


rosiemiller2:
Which bits does codecov think aren't being tested?

PR #110, opened 2020-03-17
🐛 Make env take precedence over dotenv

muchzill4:
### Description

As per title.

https://pypi.org/project/python-dotenv/
> load_dotenv does not override existing System environment variables. To override, pass override=True to load_dotenv().

### Test Coverage


mbed_targets/config.py
```diff
@@ -17,7 +17,7 @@
 import os
 import dotenv
 
-dotenv.load_dotenv(dotenv.find_dotenv(usecwd=True), override=True)
+dotenv.load_dotenv(dotenv.find_dotenv(usecwd=True))
```
rwalton-arm: Don't we need to actually read from the env at the call site? We cache the variable at import time so we could get out of sync with the user's shell environment?

muchzill4: Isn't this true of this and previous approach where variables were assigned at load time? This has not been introduced by this PR, right?

rwalton-arm: It wasn't true in the original code (i.e when this lived in `target_database`) as we read from the env in `_get_request`. Must've changed when it was moved here, and I missed it in the review.

muchzill4: Could we address that separately then? I'd be keen not to pollute this PR with unrelated bugfix. :)

rwalton-arm: Yeah, makes sense.

PR #118, opened 2020-03-25
♻ Move MbedTarget into separate module

muchzill4:
### Description

Part one of refactor, to make this easier to consume - move `MbedTarget` to a separate module.

### Test Coverage


mbed_targets/mbed_target.py
```diff
@@ -0,0 +1,95 @@
...
+            product_code=target_entry.get("product_code", ""),
+            target_type=target_entry.get("target_type", ""),
+            slug=target_entry.get("slug", ""),
+            mbed_os_support=tuple(target_entry.get("mbed_os_support", [])),
+            mbed_enabled=tuple(target_entry.get("mbed_enabled", [])),
+            build_variant=tuple(target_entry.get("build_variant", [])),
+        )
+
+
+def get_target_build_attributes(mbed_target: MbedTarget, path_to_targets_json: str) -> Any:
```
rwalton-arm: Hmm, I'm wondering if this belongs with the other public functions rather than being here with `MbedTarget` - might be easier to see what the interface to the module is if it's all in one place.

muchzill4: Define `with the other public functions`. I actually failed to group everything that takes `MbedTarget` as an argument together - planning to do it in the next step.

rwalton-arm: Oh, I mean the `get_target_by_*` functions which are the public interface to the package.

muchzill4: Hm, this has not been exposed publicly in `__init__`, so I'm not sure if this is ready to be public yet. Also, I was leaning towards having everything that takes `MbedTarget` in one place. Let's keep it where it is for the time being, unless you feel strongly otherwise. :)

rwalton-arm: I think that's fine. We can always move it later if we need to.

PR #122, opened 2020-03-27
♻️ Improve build attributes

rosiemiller2:
### Description
Refactor some of the build attributes parser code based on the feedback:

1. Don't hard-code all possible modifiers of accumulating attributes. 
2. Reuse the naming code for both `ALL_ACCUMULATING_ATTRIBUTES` and adding and removing elements in `_calculate_attribute_elements`.
3. Declare the `NON_OVERRIDING_ATTRIBUTES` at the top of the file in the same way as accumulating attributes. 

### Test Coverage


mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
```diff
@@ -13,18 +13,31 @@
...
+    """
+    all_accumulating_attributes = ACCUMULATING_ATTRIBUTES.copy()
+    for modifier_action in ACCUMULATING_ATTRIBUTES_MODIFIERS.keys():
+        all_accumulating_attributes.extend(
+            [_modifying_attribute_name(attribute, modifier_action) for attribute in ACCUMULATING_ATTRIBUTES]
+        )
+    return all_accumulating_attributes
+
+
+ALL_ACCUMULATING_ATTRIBUTES = _list_all_accumulating_attributes_and_modifiers()
```
rosiemiller2: I am not sure about this. I don't like that I have to put the private functions above this in the file, but I also don't want to just provide a function that recreates the list over and over again unnecessarily. I also want to be able to pull the constant `ALL_ACCUMULATING_ATTRIBUTES` into the overriding attributes parser to create `NON_OVERRIDING_ATTRIBUTES`.  🤔  Maybe I am taking the wrong approach. Any suggestions welcome.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
```diff
@@ -13,18 +13,31 @@
 from typing import Dict, List, Any, Deque
 
 ACCUMULATING_ATTRIBUTES = ["extra_labels", "macros", "device_has", "features", "components"]
```
rwalton-arm: Should this be a `tuple` rather than a `list` if we're saying it's a constant? Mutable constants make my head hurt :)

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
muchzill4: Less code, but unsure if readable. `itertools` is quite powerful. Maybe somebody has a better idea.
```
ALL_ACCUMULATING_ATTRIBUTES = ACCUMULATING_ATTRIBUTES + [f"{attr}{suffix}" for attr, suffix in itertools.product(ACCUMULATING_ATTRIBUTES, ["_add", "_remove"])]
```

Also, should these contants be `set`s?

muchzill4: Alternative:
```
_WITH_ADD = [f"{attr}_add" for attr in ACCUMULATING_ATTRIBUTES]
_WITH_REMOVE = ...
ALL = ACCUMULATING_ATTRIBUTES + _WITH_ADD + _WITH_REMOVE
```

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
```diff
@@ -13,18 +13,31 @@
...
-    "macros_remove",
-    "macros_add",
-    "device_has_remove",
-    "device_has_add",
-    "features_remove",
-    "features_add",
-    "components_remove",
-    "components_add",
-]
+ACCUMULATING_ATTRIBUTES_MODIFIERS = {"add": "_add", "remove": "_remove"}
```
muchzill4: Is there a reason I'm missing why this is a dict instead of a list? 🤔 
I.e.: `MODIFIERS = ["add", "remove"]`?

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rosiemiller2: In the sets v tuples I think I prefer tuples. Doesn't really matter if there are duplicates but I want it to be immutable I think.

I like the idea of using itertools. I will have a play around and see how it looks.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rosiemiller2: Only so I don't have to do a separate check that the modifying action I pass to _modifying_attribute_name is in the list, since its a dictionary it will just tell me its not there. But if this isn't a good reason I am happy to change it.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
```diff
@@ -13,18 +13,31 @@
...
+    """Provides a list of all accumulating attributes and their modifying attributes.
+
+    For example, the accumulating attribute "extra_labels" can be modified by "_add" and
+    "_remove" to make "extra_labels_add" and "extra_labels_remove".
+
+    Returns:
+        A list of all the accumulating attributes and their modifiers.
+    """
+    all_accumulating_attributes = ACCUMULATING_ATTRIBUTES.copy()
+    for modifier_action in ACCUMULATING_ATTRIBUTES_MODIFIERS.keys():
```
rwalton-arm: Not sure we need to call the `keys` here, by default the dictionary iterator should iterate through the keys.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rwalton-arm: Should we make this a class and just create the `all_accumulating_attributes` as an instance variable in the `__init__` method?

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
```diff
@@ -13,18 +13,31 @@
...
-    "device_has_add",
-    "features_remove",
-    "features_add",
-    "components_remove",
-    "components_add",
-]
+ACCUMULATING_ATTRIBUTES_MODIFIERS = {"add": "_add", "remove": "_remove"}
+
+
+def _modifying_attribute_name(attribute: str, modifying_action: str) -> str:
```
rwalton-arm: Should this be `modify_attribute_name`? Or is a "modifying attribute" a thing? We seem to pass an "accumulating attribute" to this. TBH I'm finding some of the terminology confusing so if I'm making no sense ignore me.

rosiemiller2: No its true, I'm using modifying here as an adjective and not a verb, as in you get an accumulating attribute (eg. "extra_labels") and then you get attributes that modify that accumulating attribute (eg. "extra_labels_add", "extra_labels_remove"). But yeah its confusing.  I'll have a think about better naming.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
muchzill4: You can achieve the same result with a list - `"add" in ["add", "remove"]`. 
I think this dictionary would make more sense if `key` was substantially different to `suffix`. Am I missing something? 🤔

rosiemiller2: No thats fair. I'm redoing it :)

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rosiemiller2: 45ed7ec

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rosiemiller2: 45ed7ec

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rosiemiller2: 45ed7ec

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rosiemiller2: 45ed7ec

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rosiemiller2: I'm not sure that would much help. I want to be able to access it from other files without having to initialise anything.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
rosiemiller2: I have spent hours going round in circles about naming. I have added a bit more explanation to the docstring in the hope that is helps a little.

mbed_targets/_internal/target_attribute_hierarchy_parsers/accumulating_attribute_parser.py
```diff
@@ -12,19 +12,11 @@
...
-    "device_has_remove",
-    "device_has_add",
-    "features_remove",
-    "features_add",
-    "components_remove",
-    "components_add",
-]
+ACCUMULATING_ATTRIBUTES = ("extra_labels", "macros", "device_has", "features", "components")
+MODIFIERS = ("add", "remove")
+ALL_ACCUMULATING_ATTRIBUTES = ACCUMULATING_ATTRIBUTES + tuple(
```
rwalton-arm: This is much more readable IMO. 👍

PR #125, opened 2020-03-31
♻️Separate build attributes out into a new class

rosiemiller2:
### Description
Created an `MbedTargetBuildAttributes` class and methods for accessing this with either a board type or an `MbedTarget` instance.

This class doesn't yet have all the fields needed by mbed-build, but I would rather we added them as we need them. That way we can keep track of what parts of the targets.json definition are actually being used and what are just relics of older configuration.

### Test Coverage


mbed_targets/mbed_target_build_attributes.py
```diff
@@ -0,0 +1,57 @@
...
+        supported_toolchains: Toolchains that can be used to build for this target.
+        default_toolchain: Default toolchain used for this target.
+    """
+
+    labels: FrozenSet[str]
+    features: FrozenSet[str]
+    components: FrozenSet[str]
+    config: Dict[str, str]
+    supported_toolchains: FrozenSet[str]
+    default_toolchain: str
```
muchzill4: This should be `Optional[str]` as below `get('', None)` is used.

mbed_targets/get_build_attributes.py
```diff
@@ -0,0 +1,43 @@
...
+    The build attributes are as defined in the targets.json file linked to.
+
+    Args:
+        mbed_target: the Mbed Target object representing the target data from the online database
+        path_to_targets_json: path to Mbed OS's targets.json file
+
+    Raises:
+        TargetBuildAttributesError: an error has occurred while fetching build attributes
+    """
+    return MbedTargetBuildAttributes.from_board_type(mbed_target.board_type, path_to_targets_json)
```
muchzill4:
```suggestion
    return get_build_attributes_by_board_type(mbed_target.board_type, path_to_targets_json)
```

Thoughts? **It'd need updating the test, so don't commit!**

mbed_targets/mbed_target_build_attributes.py
muchzill4: Shows how good `mypy` is, eh? ;)

rosiemiller2: efc3f58

mbed_targets/get_build_attributes.py
rosiemiller2: Yeah I like it 82cfc04.

PR #127, opened 2020-04-02
Target core labels

rosiemiller2:
### Description
Introduce the mbed_os_metadata.json file and add the core label configuration to it. 
This file should eventually be moved to be located next to targets.json.
Read out the core labels into the build attributes 'labels' field. 

### Test Coverage


mbed_targets/_internal/target_attributes.py
```diff
@@ -100,3 +105,19 @@ def _extract_target_attributes(all_targets_data: Dict[str, Any], target_name: st
     accumulated_attributes = get_accumulating_attributes_for_target(all_targets_data, target_name)
     target_attributes.update(accumulated_attributes)
     return target_attributes
+
+
+def _extract_core_labels(build_attributes: Dict[str, Any]) -> Set[str]:
```
muchzill4, line 110: Since `core_labels` relies only on `core_name` I think it'd be better if that was passed instead. That way the knowledge of how `build_attributes` is structured is moved up the call stack.

rosiemiller2: Yeah that would be much better 👍

rosiemiller2: b0b42e5

PR #133, opened 2020-04-14
♻️Renaming

rosiemiller2:
### Description

A complete adjustment of naming in mbed-targets.

What was an MbedTarget is now a Board. What was an MbedTargetBuildAttributes is now a Target. 

This rename is based on the terminology work summarised in the diagram https://excalidraw.com/#room=ee49abdc45291795d2d0,zfNDGElbWYeznYU0izj5vA and the b-tools glossary https://confluence.arm.com/display/IoTWeb/Glossary.  

It is trying to help reflect that the os site data is for development boards/platforms only and the definition of target should be narrowed to only the 'destination of built code', not the more general term for hardware (which should be 'devices').

If in the future these conditions change (eg. the online db starts to support production hardware data) then these names will have to be updated.

I have made example PRs for how these changes will affect mbed-devices (https://github.com/ARMmbed/mbed-devices/pull/60) and mbed-build (https://github.com/ARMmbed/mbed-build/pull/17).

### Test Coverage


mbed_targets/_internal/target_attributes.py
```diff
@@ -82,24 +82,25 @@ def _read_json_file(path_to_file: pathlib.Path) -> Any:
...
 
 def _extract_target_attributes(all_targets_data: Dict[str, Any], target_name: str) -> Any:
-    """Extracts the attributes for a particular target from the targets data.
+    """Extracts the definition for a particular target from all the targets in targets.json.
 
     Args:
         all_targets_data: a dictionary representation of targets.json data, still
-        containing all the hierarchy structure.
-        target_name: the name of the target also known as 'board_type' in the online database.
+        containing all the un-flattened hierarchy structure.
```
muchzill4, line 89: Maybe something like `raw targets.json data` would be a clearer description of the first argument? The fact it's a dict is clear because of typing, _un-flattened hierarchy structure_ I find hard to consume.

rosiemiller2: c979061

ci_scripts/sync_board_database.py
```diff
@@ -25,12 +25,12 @@
...
 
-from mbed_targets._internal.target_database import SNAPSHOT_FILENAME
-from mbed_targets.mbed_targets import MbedTargets
+from mbed_targets._internal.board_database import SNAPSHOT_FILENAME
+from mbed_targets.boards import Boards
 
 logger = logging.getLogger()
 
-TARGET_DATABASE_PATH = Path(
+BOARD_DATABASE_PATH = Path(
```
mark-edgeworth, line 33: Should the Path argument be 'mbed_boards' instead of 'mbed_targets'?

mbed_targets/_internal/target_attributes.py
```diff
@@ -34,46 +34,46 @@ class ParsingTargetsJSONError(TargetAttributesError):
...
+    target_attributes["labels"] = get_labels_for_target(all_targets_data, target_name).union(
+        _extract_core_labels(target_attributes.get("core", None))
     )
-    return build_attributes
+    return target_attributes
 
 
 def _read_json_file(path_to_file: pathlib.Path) -> Any:
-    """Reads the data from the targets.json file.
+    """Reads the data from a json file.
```
mark-edgeworth, line 66: Good spot!

mbed_targets/board.py
```diff
@@ -0,0 +1,76 @@
...
+            mbed_os_support=tuple(board_features.get("mbed_os_support", [])),
+            mbed_enabled=tuple(board_features.get("mbed_enabled", [])),
+            product_code=board_attrs.get("product_code", ""),
+            target_type=board_attrs.get("target_type", ""),
+            slug=board_attrs.get("slug", ""),
+            # NOTE: Currently we hard-code the build variant for a single board type.
+            # This is simply so we can demo the tools to PE. This must be removed and replaced with a proper mechanism
+            # for determining the build variant for all platforms. We probably need to add this information to the
+            # targets database.
+            build_variant=cast(Tuple, ("S", "NS") if "lpc55s69" in board_attrs.get("board_type", "") else ()),
```
mark-edgeworth, line 61: Is there a way of marking this TODO or something so that it isn't forgotten?

ci_scripts/sync_board_database.py
rosiemiller2: Ah no, here it refers to 'mbed_targets' the top level directory name rather than `mbed_targets` a variable name. This hasn't changed since the package is still `mbed-targets`.

mbed_targets/board.py
rosiemiller2: a0fa3ca

PR #136, opened 2020-04-20
✨ Parse config attribute

rosiemiller2:
### Description

In targets.json most of the overriding attributes override each other completely. Eg:

```
{
   "Target": {
          "attribute": {"description": "everything in here is overridden"},
          "attribute_2": "I stay put because I'm not overridden"
    },
    "K64F": {
          "inherits" : "Target",
          "attribute": {"another whole attribute": "has blatted the other one"}
    }
}
k64f_after_parsing = {
    "inherits": "Target",
    "attribute": {"another whole attribute": "has blatted the other one"},
    "attribute_2": "I stay put because I'm not overridden"
}
```
But the config doesn't get completely blatted by subsequent definitions, only its internal elements get overridden. Eg:

```
{
   "Target": {
          "config": {
                 "description": "everything in here is overridden",
                 "another_description": "I stay put"
           },
          "attribute_2": "I stay put because I'm not overridden"
    },
    "K64F": {
          "inherits" : "Target",
          "config": {"description": "this only blats the one element"}
    }
}
k64f_after_parsing = {
    "inherits": "Target",
    "config": {
          "description": "this only blats the one element",
          "another_description": "I stay put"
     },
    "attribute_2": "I stay put because I'm not overridden"
}
```
This code parses this additional complexity correctly
.
### Test Coverage


mbed_targets/_internal/targets_json_parsers/overriding_attribute_parser.py
```diff
@@ -93,11 +113,21 @@ def _determine_overridden_attributes(override_order: List[dict]) -> Dict[str, An
     Returns:
         A dictionary containing all the overridden attributes for a target
     """
-    target_attributes = reduce(lambda x, y: {**x, **y}, reversed(override_order))
+    target_attributes = _reduce_right_list_of_dictionaries(override_order)
+    for merging_attribute in MERGING_ATTRIBUTES:
+        override_order_for_single_attribute = map(lambda x: cast(dict, x.get(merging_attribute, {})), override_order)
```
muchzill4, line 118: I find this name a bit confusing.
`override_order` indicates to me that it is a representation of ordering rules, how about `overrides_in_order` or `ordered_overrides`?

You could also use a list comprehension here - might be easier to read:
```
result: List[dict] = [x.get(merging_attributes, {}) for x in override_order]
```

mbed_targets/_internal/targets_json_parsers/overriding_attribute_parser.py
```diff
@@ -93,11 +113,21 @@ def _determine_overridden_attributes(override_order: List[dict]) -> Dict[str, An
...
+        merged_attribute_elements = _reduce_right_list_of_dictionaries(list(override_order_for_single_attribute))
```
muchzill4, line 119: You might lose additional config option attributes here. I'm not sure what level of overrides this operates on, but isn't it possible that:
```py
source_1 = {
  "config": {
    "variable_1": {
      "help": "help text",
      "value": 123
    }
  }
}
config_2 = {
 "config": {
   "variable_1": 456
  }
}
```
Merging 2 into 1 will blat "help"? Do I understand this correctly?

muchzill4:
I approve of those ascii graphs!

mbed_targets/_internal/targets_json_parsers/overriding_attribute_parser.py
rosiemiller2: Yeah it will. Thats a good point, I had assumed that the merging was only needed at that one level but maybe its all the way down 🤔 I will check in the existing tools.

mbed_targets/_internal/targets_json_parsers/overriding_attribute_parser.py
rosiemiller2: 👍

mbed_targets/_internal/targets_json_parsers/overriding_attribute_parser.py
rosiemiller2: Yes, this is ok. There shouldn't be any merging of config params, they just overwrite.  The example you give above is actually handled with the 'overrides' field which I will make a separate PR to handle.

mbed_targets/_internal/targets_json_parsers/overriding_attribute_parser.py
rosiemiller2: Gah naming is the most difficult. Not sure I have improved it much but fwiw - 905635a and 6d51d12.

muchzill4: I think `targets_in_order` reads better than `accumulation_order`. 👍

